Issue #2, opened 2016-08-30
config-mlnx-sn2700.patch fails to apply


rspolyak:
During compilation config-mlnx-sn2700.patch patch fails:

Checking for changes in the working directory ... done
Importing patch "config-dell-s6000.patch" ... done
Importing patch "config-mlnx-sn2700.patch" ... error: patch failed: debian/build/build_amd64_none_amd64/.config:3492
error: debian/build/build_amd64_none_amd64/.config: patch does not apply
stg import: Diff does not apply cleanly

rspolyak:
In this particular case seems that config-mlnx-sn2700.patch was based on some other version of .config file than retrieved from launchpad.
CONFIG_SENSORS_CISCO_N2200=m is present in config-mlnx-sn2700.patch, but is not available in 
debian/build/build_amd64_none_amd64/.config

PR #3, opened 2016-09-13
add Cavium as7512-32x related kernel driver patches

rspolyak:
Script build.sh was extended in order to allow specifying path with vendor patches. This will help to eliminate issues like  https://github.com/Azure/sonic-linux-kernel/issues/2, when different vendors try to patch the same config file.
This change does not impact existing vendors


garyachy:
@stcheng and @lguohan please review

rspolyak:
@stcheng and @lguohan please review

lguohan:
The sonic kernel take a minimum approach, it only takes kernel patch that has to be build within the kernel tree. All the hardware platform share a same kernel.
i.e., kernel configuration changes, existing driver patches. For other loadable modules needed by the a certain platform, we have repo such as https://github.com/Azure/sonic-platform-modules-s6000.

patch/cavm-as7512/config-accton-as7512_32x.patch
```diff
@@ -0,0 +1,71 @@
...
+diff --git a/debian/build/build_amd64_none_amd64/.config b/debian/build/build_amd64_none_amd64/.config
+index c0b11a1..df010e7 100644
+--- a/debian/build/build_amd64_none_amd64/.config
++++ b/debian/build/build_amd64_none_amd64/.config
+@@ -1788,6 +1788,8 @@ CONFIG_EEPROM_LEGACY=m
+ CONFIG_EEPROM_MAX6875=m
+ CONFIG_EEPROM_93CX6=m
+ # CONFIG_EEPROM_93XX46 is not set
++CONFIG_EEPROM_ACCTON_AS7512_32x_SFP=m
++CONFIG_EEPROM_SFF_8436=m
```
lguohan: this is already in https://github.com/Azure/sonic-linux-kernel/blob/master/patch/config-dell-s6000.patch, please check

patch/cavm-as7512/driver-i2c-muxes-pca954x-always-deselect.patch
```diff
@@ -0,0 +1,23 @@
+Always deselect channel after transaction for PCA954x.
```
lguohan: Can you explain the nature of this patch, is it a driver bug we are fixing (upstreamed) or is it a workaround for the hardware platform.

lguohan:
for driver-patches-for-accton-as7512-fan-psu-cpld-led-sfp.patch, can we build out of kernel tree like the dell s6000 platform modules? We can setup repo for you to contribute.

lguohan:
3.16.7 kernel has an igb driver, what's the reason to upgrade to 5.1.2 version?

paulmenzel:
What is the status of this merge/pull request? As there hasn’t been any reply, should it be closed?

PR #10, opened 2016-12-25
update linux kernel to 3.16.36-1 deb8u2

lguohan:


lguohan:
@NStetskovych, due to multiple requests to update to latest Intel igb driver, we introduced a commit https://github.com/Azure/sonic-buildimage/pull/167 to update the intel igb driver. I did some initial test and it works with bcm54616 phy, can you please take a look at. As the result, I am removing the the patches against 5.0.5 igb driver in this repo.

NStetskovych:
@lguohan I will test

NStetskovych:
@lguohan works for me. (tested with kernel 3.16.36 and PR #167)

PR #18, opened 2017-02-28
download original files from azure storage

lguohan:


jleveque:
Build failed:

`dpkg-source: error: file ./linux_3.16.36.orig.tar.xz has size 3229396 instead of expected 81786052`

lguohan:
will merge once build finishes

lguohan:
@jleveque , fixed an issue in the previous commit.

PR #21, opened 2017-06-28
Add support for LM75B

abhisit:
I am using LM75B on my platform, name Seastone. I would like to put LM75B to the Kernel.
I am disabled file operation detect() and address_list(), since It wrong found 100+ of LM75B sensors on my machine and I would like to manual probe() by echo.

### DIFF
--- a/patch/driver-hwmon-lm75b-update.patch
+++ b/patch/driver-hwmon-lm75b-update.patch
@@ -1,0 +1,44 @@
+From 7e24a7bf79b4ae6c12cd3306f0340883b6b0fd73 Mon Sep 17 00:00:00 2001
+From: Abhisit Sangjan <asang@celestica.com>
+Date: Fri, 30 Jun 2017 10:13:12 +0700
+Subject: [PATCH] hwmon: Update to support lm75b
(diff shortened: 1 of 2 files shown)


patch/driver-hwmon-lm75b-update.patch
```diff
@@ -0,0 +1,56 @@
...
+ 	{ "max6626", max6626, },
+ 	{ "mcp980x", mcp980x, },
+@@ -465,8 +471,10 @@ static struct i2c_driver lm75_driver = {
+ 	.probe		= lm75_probe,
+ 	.remove		= lm75_remove,
+ 	.id_table	= lm75_ids,
++/*
+ 	.detect		= lm75_detect,
+ 	.address_list	= normal_i2c,
++*/
```
lguohan: we have platforms that also use lm75 sensors. Not sure if disable auto detect will affects those devices.

abhisit: Are you using x86?
How do you probe lm75 on your platform? I may follow you step, and should be work for me.

PR #22, opened 2017-06-28
Update Tun/Tap by adding enable/disable carrier ability.

VitalyVovnoboy:
Added enable/disable carrier ability in TUN/TAP driver. It used by SONiC code.


lguohan:
I am not sure if this patch is really needed or not. tun_set_iff will bring the carrier on. 

http://elixir.free-electrons.com/linux/v3.16.44/source/drivers/net/tun.c#L1696

VitalyVovnoboy:
The tun_set_iff is called when tun device is created and the device is running by default without possibility of stopping it.
We need option of setting "Running" status according to port link by controlling carrier enable/disable state.

PR #23, opened 2017-06-30
Add support for LM75B

abhisit:
I am using LM75B on my platform, name Seastone. I would like to put LM75B to the Kernel.


abhisit:
This pull-request was closed, due to this patch will be send by Celestica Account.

PR #24, opened 2017-07-03
Enable CONFIG_SCHEDSTATS for systemd-bootchart

pavel-shirshov:
we can use systemd-bootchart to understand cpu/memory/io usage. But currently it doesn't work because CONFIG_SCHEDSTATS is disabled in kernel.


lguohan:
I am not sure the kernel build uses this config file.

pavel-shirshov:
Make the change as a patch

pavel-shirshov:
This change breaks ABI.

PR #25, opened 2017-07-06
Add support for LM75B

celestica-abhisit:
The Celestica Seastone DX010 platform is using the LM75B. Please merge this commit to support LM75B.


lguohan:
can you update the sonic-linux-kernel submodule in the sonic-buildimage repo so that your kernel patches is included in the image?

PR #26, opened 2017-07-17
rtnetlink: catch -EOPNOTSUPP errors from ndo_bridge_getlink

JipanYanga:
To fix "bridge vlan show" empty result issue.

On platform which has Intel I354 nic as management port,   I354 is capable of SR-IOV,  IGB driver registered igb_ndo_bridge_getlink() and it would be called in rtnl_bridge_getlink().  But SR-IOV is not enabled by default,  igb_ndo_bridge_getlink() returns -EOPNOTSUPP which would cause rtnl_bridge_getlink() to abort the processing and returns empty or partial result to bridge vlan request


lguohan:
once it is merged, can you update the kernel submodule in the buildimage repo?

JipanYanga:
Sure, will do it.

Thanks,
Jipan

PR #27, opened 2017-08-03
Porting bridge port multicast/broadcast flood flag support, excluding netlink changes.

JipanYanga:
By default unknown unicast,  unknown multicast and broadcast packets are flooded on bridge. This change is for porting the per-port multicast flood flag and broadcasts flood flag from upstream linux to Sonic linux.


JipanYanga:
There is one line delete change in patch/series file by mistake,  closing the pr to create a new one.

PR #28, opened 2017-08-03
Porting bridge port multicast/broadcast flood flag support, excluding…

JipanYanga:
… netlink parts.

By default unknown unicast, unknown multicast and broadcast packets are flooded on bridge. This change is for porting the per-port multicast flood flag and broadcasts flood flag from upstream linux to Sonic linux.

### DIFF
--- a/patch/bridge-per-port-multicast-broadcast-flood-flags.patch
+++ b/patch/bridge-per-port-multicast-broadcast-flood-flags.patch
@@ -1,0 +1,138 @@
+commit a90be3726e45565415d1d0c04edbbb9ecd0944a1
+Author: jipan yang <jipan.yang@gmail.com>
+Date:   Tue Aug 1 23:40:26 2017 -0700
+
+    Porting bridge port multicast/broadcast flood flag support, excluding netlink parts.
+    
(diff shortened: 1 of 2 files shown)


patch/bridge-per-port-multicast-broadcast-flood-flags.patch
```diff
@@ -0,0 +1,138 @@
+commit a90be3726e45565415d1d0c04edbbb9ecd0944a1
+Author: jipan yang <jipan.yang@gmail.com>
+Date:   Tue Aug 1 23:40:26 2017 -0700
+
+    Porting bridge port multicast/broadcast flood flag support, excluding netlink parts.
```
lguohan: why excluding netlink parts?

lguohan: so, you only plan to use sysfs to control this?

JipanYanga: Yes, sysfs is to be used to control the flooding behavior of bridge port.  

Ideally the netlink part should be back ported to sonic linux too. Once the netlink kernel part is changed, libnl library/Iproute2 and other possible netlink users should support parsing the new attributes to avoid potential issues.  Probably it is not worth all the troubles since sysfs interface is sufficient for the management purpose.

PR #31, opened 2017-09-14
Add kernel patches: fix base_reachable_time(_ms) not effective immediatly when changed

gongjianLhr:
Signed-off-by: Haoran Liu gongjian.lhr@alibaba-inc.com


lguohan:
@gongjianLhr, the kernel has been upgraded to 4.9 kernel, is this patch still needed?

PR #32, opened 2017-11-20
i2c: core: fix NULL pointer dereference under race condition

yurypm:
Patch from linux kernel 3.16.40
commit https://github.com/torvalds/linux/commit/147b36d5b70c083cc76770c47d60b347e8eaf231


lguohan:
@yurypm ,  can you check #33  has fixed your problem or not? I verified that the new debian kernel contains your patch.

yurypm:
@lguohan , yes #33 fixes our problem.

lguohan:
@yurypm , thanks for your confirmation. closing this PR now.

PR #33, opened 2017-11-21
upgrade kernel to 3.16.43-2+deb8u5 based on latest debian security pa…

lguohan:
…tches


stcheng:
any reasons about the kernel upgrade?
what're the features we want

SAI-shared-user:
check this PR. https://github.com/Azure/sonic-linux-kernel/pull/32

Instead of pulling the a specific commit, I think it is better to upgrade the kernel image to include latest security bug fixes.

PR #37, opened 2017-12-30
Disable suspend functionality

pavel-shirshov:
Otherwise some devices could be suspended and then woke up with different names.

### DIFF
--- a/patch/config-disable-suspend.patch
+++ b/patch/config-disable-suspend.patch
@@ -1,0 +1,13 @@
+diff --git a/debian/build/build_amd64_none_amd64/.config b/debian/build/build_amd64_none_amd64/.config
+index 4efabb8..587cef3 100644
+--- a/debian/build/build_amd64_none_amd64/.config
++++ b/debian/build/build_amd64_none_amd64/.config
+@@ -546,7 +546,7 @@ CONFIG_USE_PERCPU_NUMA_NODE_ID=y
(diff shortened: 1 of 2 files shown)


patch/config-disable-suspend.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/debian/build/build_amd64_none_amd64/.config b/debian/build/build_amd64_none_amd64/.config
+index 4efabb8..587cef3 100644
+--- a/debian/build/build_amd64_none_amd64/.config
++++ b/debian/build/build_amd64_none_amd64/.config
+@@ -546,7 +546,7 @@ CONFIG_USE_PERCPU_NUMA_NODE_ID=y
+ # Power management and ACPI options
+ #
+ CONFIG_ARCH_HIBERNATION_HEADER=y
+-CONFIG_SUSPEND=y
```
qiluo-msft: Set it to 'n'?

pavel-shirshov: It's a Linux standard way to make the config option "n" by commenting it out. I think this is because some old kernel code still can use ```#ifdef CONFIG_OPTION```

pavel-shirshov:
I'll fix this PR ASAP. I'm checking it again now.

pavel-shirshov:
This change breaks kernel ABI, so I decided to disable this functionality using kernel command line options

PR #41, opened 2018-02-09
drivers/misc/eeprom:  Add support for optoe driver

donboll:
* Add the optoe driver, copied from github.com/opencomputeproject/oom
* Update Makefile to build optoe
* Update Kconfig to conifuge optoe

Signed-off-by: Don Bollinger <don@thebollingers.org>


lguohan:
I am not able to find the optoe driver compiled in the kernel build log. Can you check?

https://sonic-jenkins.westus2.cloudapp.azure.com/job/common/job/linux-kernel-pr-build/58/consoleText

lguohan:
I think what is missing here is probably the kernel config, you need to enable to compile your driver as a kernel module. check this patch for sff_8436 driver.

https://github.com/Azure/sonic-linux-kernel/blob/master/patch/config-dell-s6000.patch#L45

PR #42, opened 2018-03-05
add ipmi and lpc in the kernel

hans-tseng:
Signed-off-by: hans <hans.tseng@deltaww.com>
add ipmi and lpc in the kernel which delta_ag9064 used

### DIFF
--- a/patch/config-delta-ag9064.patch
+++ b/patch/config-delta-ag9064.patch
@@ -1,0 +1,44 @@
+diff --git a/debian/build/build_amd64_none_amd64/.config b/debian/build/build_amd64_none_amd64/.config
+index e9b840e..dea8b32 100644
+--- a/debian/build/build_amd64_none_amd64/.config
++++ b/debian/build/build_amd64_none_amd64/.config
+@@ -3090,13 +3090,13 @@ CONFIG_HVC_IRQ=y
+ CONFIG_HVC_XEN=y
+ CONFIG_HVC_XEN_FRONTEND=y
+ CONFIG_VIRTIO_CONSOLE=m
+-CONFIG_IPMI_HANDLER=m
++CONFIG_IPMI_HANDLER=y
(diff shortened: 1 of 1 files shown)


patch/config-delta-ag9064.patch
```diff
@@ -0,0 +1,44 @@
+diff --git a/debian/build/build_amd64_none_amd64/.config b/debian/build/build_amd64_none_amd64/.config
+index e9b840e..dea8b32 100644
+--- a/debian/build/build_amd64_none_amd64/.config
++++ b/debian/build/build_amd64_none_amd64/.config
+@@ -3090,13 +3090,13 @@ CONFIG_HVC_IRQ=y
+ CONFIG_HVC_XEN=y
+ CONFIG_HVC_XEN_FRONTEND=y
+ CONFIG_VIRTIO_CONSOLE=m
+-CONFIG_IPMI_HANDLER=m
++CONFIG_IPMI_HANDLER=y
```
lguohan: why do we need to change to built-in option? is kernel module ok?

lguohan:
not sure we need to build all those module into kernel, can we leave them as kernel module?

patch/config-delta-ag9064.patch
hans-tseng: Hi, sir:
    Because to load the LPC and ipmi  module has some timing and sequence issue and they should be put into kernel without loadable. 

Hans

hans-tseng:
Hi, sir:
Because to load the LPC and ipmi module has some timing and sequence issue and they should be put into kernel without loadable.

Hans

lguohan:
what kind of sequencing issues are you facing? can you share details?

hans-tseng:
Hi Sir:

    I think we had found the way to use the loadable module, 
    I will close this request and update the loadable command in the ag9064 of biuld0image

Hans

PR #43, opened 2018-03-06
drivers/misc/eeprom: optoe patch - improve EOF handling for access beyond end of device

donboll:


donboll:
Roy Lee at Accton discovered in his testing that attempts to read beyond the end of devices that do not page results in unexpected error messages.  With this patch optoe returns '0' for accesses that start beyond the end of the device (256 bytes for a DAC or a QSFP without pages, 512 bytes for an SFP without pages, 32K+ for paged devices).  According to the Linux Device Driver book (LDD3, ch3, page 66), "The return value for read is interpreted by the calling application program...  If the value is 0, end-of-file was reached (and no data was read)."  This patch implements this policy.  Roy has confirmed the unexpected error messages are now gone.

donboll:
I seem to have done something wrong in the pull request.  The only change I am requesting is the last commit, f88d7ab, the patch to improve optoe handling of EOF.  Please advise if I need to change this PR.

Don

lguohan:
@donboll, you do not need all those kernel packages, and you need to put the patch into series

donboll:
Please disregard this pull request.  I will send a cleaner PR with the same changes momentarily.

PR #44, opened 2018-03-11
Update to optoe driver to improve EOF handling

donboll:
Added patch: 'driver-support-optoe-EOF_fix.patch' to patch directory
	Added the patch to 'series'

Signed-off-by: Don Bollinger

### DIFF
--- a/patch/driver-support-optoe-EOF_fix.patch
+++ b/patch/driver-support-optoe-EOF_fix.patch
@@ -1,0 +1,149 @@
+From 556c452dcf5706a951ef490327b74bd27a411693 Mon Sep 17 00:00:00 2001
+From: Don Bollinger <don@thebollingers.org>
+Date: Fri, 9 Mar 2018 18:22:33 -0800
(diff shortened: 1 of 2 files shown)


donboll:
Please accept this PR to replace PR #43.

Roy Lee at Accton discovered in his testing that attempts to read beyond the end of devices that do not page results in unexpected error messages. With this patch optoe returns '0' for accesses that start beyond the end of the device (256 bytes for a DAC or a QSFP without pages, 512 bytes for an SFP without pages, 32K+ for paged devices). According to the Linux Device Driver book (LDD3, ch3, page 66), "The return value for read is interpreted by the calling application program... If the value is 0, end-of-file was reached (and no data was read)." This patch implements this policy. Roy has confirmed the unexpected error messages are now gone.

This PR is implemented as a new patch in the 'patch' directory, which will patch optoe.c.  It also updates the 'series' file to reference this patch, after the patch that introduces optoe.c.  

Don

PR #47, opened 2018-04-18
Add NVMe support using for QUANTA IX4 switch (#47)

jonathantsai-qci:
* Add NVMe support using for QUANTA switch

Signed-off-by: Jonathan Tsai <jonathan.tsai@quantatw.com>

### DIFF
--- a/patch/config-quanta-ix4.patch
+++ b/patch/config-quanta-ix4.patch
@@ -1,0 +1,13 @@
+diff --git a/debian/build/build_amd64_none_amd64/.config b/debian/build/build_amd64_none_amd64/.config
+index 6bd2b13..9292a16 100644
+--- a/debian/build/build_amd64_none_amd64/.config
++++ b/debian/build/build_amd64_none_amd64/.config
+@@ -1723,7 +1723,7 @@ CONFIG_BLK_DEV_LOOP_MIN_COUNT=8
+ CONFIG_BLK_DEV_DRBD=m
+ # CONFIG_DRBD_FAULT_INJECTION is not set
+ CONFIG_BLK_DEV_NBD=m
+-CONFIG_BLK_DEV_NVME=m
++CONFIG_BLK_DEV_NVME=y
+ CONFIG_BLK_DEV_SKD=m
+ CONFIG_BLK_DEV_OSD=m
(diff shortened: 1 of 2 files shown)


patch/config-quanta-ix4.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/debian/build/build_amd64_none_amd64/.config b/debian/build/build_amd64_none_amd64/.config
+index 6bd2b13..9292a16 100644
+--- a/debian/build/build_amd64_none_amd64/.config
++++ b/debian/build/build_amd64_none_amd64/.config
+@@ -1723,7 +1723,7 @@ CONFIG_BLK_DEV_LOOP_MIN_COUNT=8
+ CONFIG_BLK_DEV_DRBD=m
+ # CONFIG_DRBD_FAULT_INJECTION is not set
+ CONFIG_BLK_DEV_NBD=m
+-CONFIG_BLK_DEV_NVME=m
++CONFIG_BLK_DEV_NVME=y
```
lguohan: why needs built-in? why cannot use loadable module?

jonathantsai-qci: image
hi guohan

our kernel need to put under nvme but not flash, and so that we also have to changed install.sh.
please check above picture, thanks.

cpchangtw: Hi @lguohan 
Any concern for this ? maybe u got another better way to acheive this goal? thanks.

lguohan: change install.sh is total separate from kernel change.

why do you need to compile nvme as built-in instead of a kernel module. I do not understand the logic here.

cpchangtw: @lguohan 
IX4 uses PCIe NVME for install kernel image, not PCIe mSATA.
More specifically, the device should be "nvme0n1" but not "mmcblk0".
Any advice to achieve same goal with nvme as kernel module?

cpchangtw: @lguohan Any advise?

jonathantsai-qci:
Since nvme loaded in initramfs need time to verified, close this temporally.

PR #49, opened 2018-06-27
Support Delta DPS1900

deni64k:
Hi there,

Here is an implementation of the module dps1900 to support Delta DPS1900 devices.

Thanks for your review.


patch/driver-hwmon-pmbus-dps1900.patch
```diff
@@ -0,0 +1,104 @@
...
++		.name = "dps1900",
++	},
++	.probe = dps1900_probe,
++	.remove = pmbus_do_remove,
++	.id_table = dps1900_id,
++};
++
++module_i2c_driver(dps1900_driver);
++
++MODULE_AUTHOR("Arista Networks");
```
lguohan: arista networks?

deni64k:
Thank you, @lguohan!

PR #51, opened 2018-07-02
[Kernel] support downloading built kernel Debian packages

yxieca:
Add support to download pre-built Linux kernel Debian packages instead of building from source files.

This change prefers the method of download over build.

Note: this PR is better go in after https://github.com/Azure/sonic-buildimage/pull/1837 but it doesn't have to.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,38 @@
 .ONESHELL:
 SHELL = /bin/bash
 
+ifneq ($(kernel_build_method), "build")
+# Downloading kernel
+
+LINUX_HEADER_COMMON = linux-headers-3.16.0-5-common_3.16.51-3+deb8u1_amd64.deb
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1,6 +1,38 @@
 .ONESHELL:
 SHELL = /bin/bash
 
+ifneq ($(kernel_build_method), "build")
```
jleveque: How does one set `kernel_build_method`? Should this become a command-line parameter specified in rules/config?

yxieca: It is. Please take a look at the linked PR.

jleveque: Aha. Sorry, missed the linked PR.

A small concern is the terminology might not be intuitive (i.e., that one of the "build methods" is "download" which doesn't actually build anything). I think a name like "kernel_procure_method" is more accurate. Maybe someone else has another suggestion?

yxieca: procure is good. Thanks.

PR #52, opened 2018-07-17
[Kernel] make sure kernel build fail on all errors and address git am error

yxieca:
Before this change, git am command (https://github.com/Azure/sonic-linux-kernel/blob/master/Makefile#L37) fails without detection.

With the first commit, kernel build fails at above location.

With the second commit. kernel build passes.


Makefile
```diff
@@ -1,5 +1,6 @@
 .ONESHELL:
 SHELL = /bin/bash
+.SHELLFLAGS = -e -c
```
qiluo-msft: >SHELLFLAGS [](start = 1, length = 10)

I think below will work:
```
.SHELLFLAGS += -e
```

Makefile
```diff
@@ -32,6 +33,7 @@ $(addprefix $(DEST)/, $(MAIN_TARGET)): $(DEST)/% :
 	pushd $(BUILD_DIR)
 	git init
 	git add -f *
+	git commit -m "check in all lose files and diffs"
```
qiluo-msft: >check in all lose files and diffs [](start = 16, length = 33)

init commit
'lose' is confusing.

PR #53, opened 2018-08-09
[Kernel] use stg to manage changelog patches

yxieca:

Managing changelog patches by stg as well. This enables multiple changelog patches.


patch/changelog/series
```diff
@@ -0,0 +1 @@
+changelog.patch
```
qiluo-msft: Why not write this line inside patch/series?

yxieca: The 2 sets of patches needs to be taken in 2 different time. We change configuration file in between.

qiluo-msft: I see. I am not familiar with stg. If you use quilt, you can use 'quilt push' to control the phases.

yxieca: I looked briefly at guilt, it manages a serial of patches like stg, but stg is working with git to create commits from each patch. That is one step further than quilt. Since we are using git, and we manage our patches with stg in other places already. I think we should stick with stg. Also, in this make file, it already uses stg to manage the rest of the patches.

yxieca: I see. Good question.

The patching has to take 2 stages, the first stage is pre-config patching, these changes affects configuration generation. Then we generate configuration with these changes. After that, the second wave of patches will be applied, some of them changes the newly generated configuration file.

This is why the patching has to take 2 stages. As we discussed offline, I changed the folder name to 'preconfig' to make it clearer.

Cheers,
Ying

PR #57, opened 2018-08-20
Fix calculation of page_len in sff_8436_read_write function

yurypm:

PR #60, opened 2018-09-07
import upstream optoe chunk offset fix

lguohan:
Correct a panic inducing defect which is triggered on a read (or write)

which begins beyond byte 128, on a byte which is not aligned on a 128 byte
chunk boundary, and which crosses a chunk boundary.

Signed-off-by: Guohan Lu <gulv@microsoft.com>


lguohan:
this is essentially the same fix as #57

donboll:
Thanks for picking  up this fix so quickly!  You're right, it is mathematically the same fix as #57.  The defect was actually inherited from sff_8436 (the ancestor of optoe).

PR #70, opened 2018-10-19
[sonic-linux-kernel] Abiname change

vharish02:
This commit adds the udp_l3mdev_accept kernel patch
to the linux kernel. This requires an abiname change
from 4.9.0-7-amd64 to 4.9.0-7-1-amd64

Signed-off-by: Harish Venkatraman <Harish_Venkatraman@dell.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@
 SHELL = /bin/bash
 .SHELLFLAGS += -e
 
-KERNEL_ABI_MINOR_VERSION = 1
+KERNEL_ABI_MINOR_VERSION = 2
 KVERSION_SHORT ?= 4.9.0-8-$(KERNEL_ABI_MINOR_VERSION)
 KVERSION ?= $(KVERSION_SHORT)-amd64
 KERNEL_VERSION ?= 4.9.110
@@ -62,8 +62,6 @@ $(addprefix $(DEST)/, $(MAIN_TARGET)): $(DEST)/% :
 	dpkg-source -x $(DSC_FILE)
 
 	pushd $(BUILD_DIR)
-	sed -i '/abiname:.*/ s/$$/-1/' debian/config/defines
-	fakeroot debian/rules debian/control-real || true
 	git init
 	git add -f *
 	git commit -qm "check in all loose files and diffs"
--- a/Makefile
+++ b/Makefile
@@ -20,11 +20,11 @@ ifneq ($(kernel_procure_method), build)
 # Downloading kernel
 
 # TBD, need upload the new kernel packages
-LINUX_HEADER_COMMON_URL = "https://sonicstorage.blob.core.windows.net/packages/kernel-public/linux-headers-4.9.0-8-common_4.9.110-3+deb9u2_all.deb?sv=2015-04-05&sr=b&sig=LlKqKecY6MDSwFMTxpKErh7FX1Lrvse2yVt3niYnhds%3D&se=2128-02-24T04%3A47%3A48Z&sp=r"
+LINUX_HEADER_COMMON_URL = "https://sonicstorage.blob.core.windows.net/packages/kernel-public/linux-headers-$(KVERSION_SHORT)-common_$(KERNEL_VERSION)-$(KERNEL_SUBVERSION)_all.deb?sv=2015-04-05&sr=b&sig=LlKqKecY6MDSwFMTxpKErh7FX1Lrvse2yVt3niYnhds%3D&se=2128-02-24T04%3A47%3A48Z&sp=r"
 
-LINUX_HEADER_AMD64_URL = "https://sonicstorage.blob.core.windows.net/packages/kernel-public/linux-headers-4.9.0-8-amd64_4.9.110-3+deb9u2_amd64.deb?sv=2015-04-05&sr=b&sig=SfQLjiBPPMcRUOmBPvXq2%2F6SsW3ul9%2FlaROplXdGij0%3D&se=2128-02-24T04%3A48%3A38Z&sp=r"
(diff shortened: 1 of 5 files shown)


patch/0025-net-udp_l3mdev_accept-support.patch
```diff
@@ -0,0 +1,279 @@
+From 58be55eb5de3a319c0740ba8071440d972c821c3 Mon Sep 17 00:00:00 2001
+From: Harish Venkatraman <Harish_Venkatraman@dell.com>
+Date: Tue, 25 Sep 2018 09:56:25 -0700
+Subject: [PATCH] net: udp_l3mdev_accept support
```
lguohan: is this upstream patch? can you put upstream commit it in the commit message?

Makefile
```diff
@@ -60,6 +60,8 @@ $(addprefix $(DEST)/, $(MAIN_TARGET)): $(DEST)/% :
 	dpkg-source -x $(DSC_FILE)
 
 	pushd $(BUILD_DIR)
+	sed -i '/abiname:.*/ s/$$/-1/' debian/config/defines
+	fakeroot debian/rules debian/control-real || true
```
lguohan: why make it always true here?

zhenggen-xu:
We are planning to move to 4.9.0-8 soon for security fixes, would this fix be compatible with that version?  Since this change touch the key modules in the kernel, it would to good to upstream it first if not yet done or share the upstream link if already exists.

Makefile
vharish02: When "fakeroot" command is run it displays a message saying "the message is intentional please ignore and continue" so true is required to continue.

patch/0025-net-udp_l3mdev_accept-support.patch
vharish02: This is a upstream patch which was back ported to 4.9 kernel manually since there is no patch available for 4.9.

> We are planning to move to 4.9.0-8 soon for security fixes, would this fix be compatible with that version? Since this change touch the key modules in the kernel, it would to good to upstream it first if not yet done or share the upstream link if already exists.

This is a patch which is back ported to 4.9 for SONiC. Secondly this fix is self contained for UDP and should be compatible.

Makefile
vharish02: When "fakeroot" command is run it displays a message saying "the message is intentional please ignore and continue" so true is required to continue.

qiluo-msft:
The download links don't work because there are not available there.

lguohan:
please describe the test you have done.

Makefile
```diff
@@ -2,7 +2,7 @@
 SHELL = /bin/bash
 .SHELLFLAGS += -e
 
-KVERSION_SHORT ?= 4.9.0-7
+KVERSION_SHORT ?= 4.9.0-7-1
```
lguohan: please define a macro for abi minor version and use it throughout the Makefile.

patch/0025-net-udp_l3mdev_accept-support.patch
jipanyang: I guess the changes you put here originate from few upstream commits after 4.9 though they may not be a exact match.  Could you you list  those upstream commits in your commit message?

Makefile
```diff
@@ -60,6 +60,8 @@ $(addprefix $(DEST)/, $(MAIN_TARGET)): $(DEST)/% :
 	dpkg-source -x $(DSC_FILE)
 
 	pushd $(BUILD_DIR)
+	sed -i '/abiname:.*/ s/$$/-1/' debian/config/defines
```
jipanyang: I remember for debian kernel build, if there is any change with kernel/user ABI,  some ABI function file has to be updated explicitly.   Did you change any of the ABI files?

Makefile
vharish02: Guohan I have addressed your comments and updated the PR. Created a new macro for abi minor version and also changed the URL link accordingly in the Makefile

> please describe the test you have done.

I Have tested the kernel udp l3mdev_accept changes and it works. I ran a Udp server on the default VRF which binds to all IP addresses and port 10000 on the DUT and waiting for message from client. The Udp client runs on a server and sends message to DUT. 
1. Initially the udp_l3mdev_accept value is set to 0. Start the Udp server on the DUT, it is waiting for client message/data. Client is running on server,  It sends message/data to the server with mgmt-ip of DUT and port 10000. The message is not received on the DUT.  Ran multiple times.
2. I set the udp_l3mdev_accept to 1 and perform the same test. Start the server and when the client sends the message/data to mgmt-ip of DUT and port 10000 it is received on the DUT. Setting the udp_l3mdev_accept to 1 make the server running in default VRF vrf aware and receive messages from mgmt-vrf as well.

lguohan:
@vharish02 , thanks for the answer, there are other question raised a few other reviewers, can you answer those too?

vharish02:
> The download links don't work because there are not available there.
Yes That's correct currently it won't work since it is not available but Iam not sure how we store it in the location. Once we store it to the location the link should work isn't?

Makefile
vharish02: Not that I am aware of. I did not make any change explicitly.

vharish02:
> @vharish02 , thanks for the answer, there are other question raised a few other reviewers, can you answer those too?
Guohan, I have replied to that too.

lguohan:
@vharish02 , I guess the changes you put here originate from few upstream commits after 4.9 though they may not be a exact match. Could you you list those upstream commits in your commit message?

patch/0025-net-udp_l3mdev_accept-support.patch
vharish02: @jipanyang I have updated the Original commit message with the changes

Makefile
lguohan: @jipanyang , check my patch and change the abi to 8-2

PR #82, opened 2019-05-02
Fix false alarm when writing convrate on max6658

byu343:
We found that max6658 sporadically sends a false overtemp alarm during
changing the convrate by hwmon kernel driver. This workaround will fix it by 
forcing to stop conversion before setting the convrate. The false alarm will
trigger reboot and we found it on Arista 7170 switch.


lguohan:
is this a upstream fix? is this a generic fix?

byu343:
> is this a upstream fix? is this a generic fix?
It is not from upstream. 

I was thinking that the change would be generic, as it only adds some steps to change the stop/run bit of LM90_REG_R_CONFIG1, which is similar to what has been done in the function lm90_init_client(). But considering that we are not able to test the change on all the chips supported by lm90, maybe it is better to add some if-conditions to make the change only be applied to max6658. I'll make a change.

lguohan:
agree.

PR #83, opened 2019-05-15
Add ixgbe driver patch to support BCM54616S_E_PHY phy

jostar-yang:
1. Add ixgbe driver patch to support BCM54616S_E_PHY phy
2. Test mgt port networking is ok after add patch


lguohan:
we have this patch. 0011-support-Broadcom-54616-Phy-for-Intel-igb-driver.patch

what is the difference here?

lguohan:
no activity

jostar-yang:
0011-support-Broadcom-54616-Phy-for-Intel-igb-driver.patch is for IGB drvier.
IGB support Intel(R) PRO/10GbE family of adapters for PCI-X type
The patch that I add that support IXGBE. It is  Intel(R) 10GbE PCI Express family.
Intel ethernet driver put igb and ixgbe to different code.

PR #86, opened 2019-06-04
[security] Upgrade kernel to 4.9.168-1+deb9u2

qiluo-msft:


qiluo-msft:
@lguohan Could you help port the uart patches to the new kernel?

lguohan:
I have removed the two uart patch as they are already in upstream linux 4.9.168 kernel.

PR #93, opened 2019-07-25
[Mellanox] Add kernel patch to remove qsfp status caching time

keboliu:
What I did:  
In some case, after SFP plug in/out the presence status only change after about 120s, this is because of the qsfp caching time set in the kernel. To make the qsfp presence time change in time, we remove the caching time from the kernel.


qiluo-msft:
1. Why not merge the patch into patch/0006-Mellanox-switch-drivers-changes.patch?
2. Why not PR on master first, and then cherry-pick into 201811?

keboliu:
> 1. Why not merge the patch into patch/0006-Mellanox-switch-drivers-changes.patch?
> 2. Why not PR on master first, and then cherry-pick into 201811?

we don't need this patch on mater since we don't rely on the sysFS to detect the SFP presence status, we use ethtool on the master.

PR #94, opened 2019-07-31
[sflow]: backport psample and act_sample drivers, needed for sonic-vs platform

rakeshdatta:
- What I did

Backport psample and act_sample drivers to sonic linux kernel  4.9.

- How I did it

Both psample and act_sample are needed for sflow feature. 

psample driver is needed for collecting samples from the interfaces at the kernel level.
act_sample driver is needed for the 'tc' command to be able to program the sampling configurations into the interfaces at the kernel level.  

psample and act_sample drivers are not back-ported  yet to the linux kernel version that sonic-linux-kernel is using. Hence, I have taken the patches for both the drivers and included it in our sonic kernel.

- How to verify it

  - Using Azure/sonic-buildimage#3251 changes, build a sonic image
This step would include the psample and act_sample drivers into the sonic-linux-kernel.
Also, it would install the psample.ko and act_sample.ko in a sonic-vs target at the start-up time.

  - Using Azure/sonic-utilities#592 changes, use the CLIs to configure sflow globally and on a port. 
  - Thereafter, use the show CLI to verify the config changes made. 
  - Pump traffic into the port where sampling is configured. 
  - Kernel would generate NETLINK messages, containing the packet samples, and send them to the hsflowd daemon. At hsflowd side, we would be able to see the incoming packet samples.

- Description for the changelog

   - Update: Added psample driver into the sonic kernel, which enables
                   packet sampling in the vs environment
   - Update: Added act_sample driver into the sonic kernel, which enables 'sample' action
                   for the 'tc' command

Signed-off-by: Rakesh Datta <rakesh.datta@dell.com>

### DIFF
--- a/patch/kernel-enable-psample-and-act_sample-drivers.patch
+++ b/patch/kernel-enable-psample-and-act_sample-drivers.patch
@@ -1,0 +1,27 @@
+From: Rakesh Datta <rakesh_datta@dell.com>
+Date: Thu, 27 Jun 2019 11:07:08 +0100
+Subject: [PATCH] net: Introduce psample and sample modules - adding the below config:
+
+CONFIG_PSAMPLE=m
+CONFIG_NET_ACT_SAMPLE=m
+
+Signed-off-by: Rakesh Datta <rakesh_datta@dell.com>
+---
+ debian/build/build_amd64_none_amd64/.config | 2 +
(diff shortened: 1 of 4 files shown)


prsunny:
@marian-pritsak to review

PR #100, opened 2019-09-16
Added support in the kernel for fullcone 3-tuple unique nat.

kirankella:
Changes done in the kernel to ensure 3-tuple uniqueness of the conntrack entries for the fullcone nat functionality.
- Hashlist is maintained for the 3-tuple unique keys (Protocol/Source IP/Port) for all the conntrack entries.
- When NAT table rules are created with the fullcone option, the SNAT/POSTROUTING stage ensures the ports from the pool are picked up in such a way that the 3-tuple is uniquely assigned.
- In the DNAT/POSTROUTING stage, the fullcone behavior is ensured by checking and reusing the 3-tuple for the Source IP/Port in the original direction.
- When the pool is exhausted of the 3-tuple assignments, the packets are dropped, else, they will be going out of the router they being 5-tuple unique (which is not intended).

- Passing fullcone option using iptables is part of another PR (in sonic-buildimage repo).

The kernel changes mentioned above are done to counter the challenges explained in the section "3.4.2.1 Handling NAT model mismatch between the ASIC and the Kernel" in the NAT HLD.

Link to NAT HLD: 
https://github.com/kirankella/SONiC/blob/nat_doc_changes/doc/nat/nat_design_spec.md

Signed-off-by: kiran.kella@broadcom.com

### DIFF
--- a/patch/Support-for-fullcone-nat.patch
+++ b/patch/Support-for-fullcone-nat.patch
@@ -1,0 +1,693 @@
+From 67cc2bd46c7ad1431ddd1819cdfad6ecc08a7593 Mon Sep 17 00:00:00 2001
+From: Kiran Kella <kiran.kella@broadcom.com>
+Date: Fri, 6 Sep 2019 20:54:19 -0700
+Subject: [PATCH] Support for fullcone nat
+
+---
+ include/net/netfilter/nf_conntrack.h     |   4 +
+ include/net/netfilter/nf_nat.h           |   6 +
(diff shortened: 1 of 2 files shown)


lguohan:
will this get into linux kernel main stream?

kirankella:
retest this please

arlakshm:
@AkhileshSamineni, @kirankella,
Please resolve the conflicts.

PR #101, opened 2019-09-23
IPv6 host and subnet anycast route in VRF to be configured with nhop interface as actual netdev instead of VRF L3MDEV.

preetham-singh:
When we configure IPv6 route with plen <127(127 is used for inter router links), kernel auto-generates subnet-router anycast address. This is as
per RFC 4291(section Required Anycast address) and RFC 6164. In kernel
version 4.9.x these routes are autogenerated with nhop interface as VRF
L3MDEV instead of the interface on which these address are configured.
Also host address(/128) were also pointing to same L3MDEV.

This was creating issue esp during warm-reboot since FRR installs these
routes via fpm to APP_DB ROUTE_TABLE pointing to VRF name. Warm boot
will be blocked since routerorch will keep waiting for these routes to
be uninstalled.
Also these routes will not be installed to hardware since nexthop ifname
is L3MDEV.

Backporting patch from commit:
https://github.com/torvalds/linux/commit/4832c30d5458387ff2533ff66fbde26ad8bb5a2d#diff-f856f42c9bd7ceebf824b5ccadc83280


lguohan:
in the description, there is only one commit, but in the patch , there are multiple commits backported. the patch does not match the description. am i missing something here?

patch/0000-net-Fix-netdev-adjacency-tracking.patch
```diff
@@ -0,0 +1,1302 @@
...
+https://github.com/torvalds/linux/commit/790510d99f39cee7f275d001aa5024032ed9bb48#diff-3ce2f1b11ea5a4b9c88ef0b4159bd481
+
+https://github.com/torvalds/linux/commit/453d39329ad03f9f6d93ec6d0d57bf7a1e2a20c7#diff-66286e176927fcbf186d37fce71bc94e
+https://github.com/torvalds/linux/commit/e0e79c8e74b08976d9b45e52b704b9228c6965c7#diff-598fcd5f9f85a0fea3a53a7022974457
+
+https://github.com/torvalds/linux/commit/b3208b2024c9089106df52ae25ebf39068d6f9fc#diff-ce347f03cd1a8602476d0774dd8dde24
+
+https://github.com/torvalds/linux/commit/1cd127fc7d3a6d6a0fc9f9cca47ca3a16ee79679#diff-69464fa7779eabc57449a324cd70bbcb
+
+https://github.com/torvalds/linux/commit/dd82364c3ab93a96bb1e45d22106a4b1ea4bef39#diff-5a740fc1e7dea601fffed2ecbbb8e7c8
```
lguohan: https://github.com/torvalds/linux/commit/dd82364c3ab93a96bb1e45d22106a4b1ea4bef39#diff-5a740fc1e7dea601fffed2ecbbb8e7c8


according to the description, it is not related to the ipv6 nexhop feature. why is this needed?

preetham-singh: @lguohan As part of this PR there are 2 commits for 2 issues:
1. IPv6 host and subnet anycast route in VRF to be configured with nhop interface as actual netdev instead of VRF L3MDEV : Commit : https://github.com/Azure/sonic-linux-kernel/pull/101/commits/f6dab4a42bd128bb2bbe75ab9f067456b70b1389
2. Fix for Unregister_netdev VRF obsreved and kernel hangs during vrf delete : https://github.com/Azure/sonic-linux-kernel/pull/101/commits/09715a31b6f7948542c1fd5327d4d424ab3f2970

1st commit is fix for IPv6 host and subnet anycast route
2nd commit is fix for Kernel hang observed during vrf deletion. 2nd commit comprises single patch which is merged from multiple commits made to linux kernel.

Accidentally I created single PR for both issues. If required, I can create separate PR for above 2 issues. Please let me know.
Note: Both above commits are necessary for VRF feature.

lguohan: yes, please separate. thanks.

lguohan:
obsoleted by #109 and #108

PR #112, opened 2019-11-14
[kernel] add patch for mlx-platform: Fix parent device in i2c-mux-reg device registration

mykolaf:
Patch fixes parent device parameter in a call to
platform_device_register_resndata() for "i2c-mux-reg".

It solves the race during initialization flow while 'i2c_mlxcpld.1' is
removing after probe, while 'i2c-mux-reg.0' is still in probing flow.

Signed-off-by: Mykola Faryma <mykolaf@mellanox.com>


qiluo-msft:
Why this is not submitted against master branch?

PR #116, opened 2019-12-15
Security kernel update to 4.9.189

olivier-singla:
This patch upgrade the kernel from version
4.9.0-9-2 (4.9.168-1+deb9u3) to 4.9.0-11-2 (4.9.189-3+deb9u2)

This PR needs another PR in sonic-buildimage:
https://github.com/Azure/sonic-buildimage/pull/3913


Makefile
```diff
@@ -2,11 +2,11 @@
 SHELL = /bin/bash
 .SHELLFLAGS += -e
 
-KERNEL_ABI_MINOR_VERSION = 2
-KVERSION_SHORT ?= 4.9.0-9-$(KERNEL_ABI_MINOR_VERSION)
+KERNEL_ABI_MINOR_VERSION = 11-2
```
qiluo-msft: >KERNEL_ABI_MINOR_VERSION [](start = 0, length = 24)

No need to change KERNEL_ABI_MINOR_VERSION, it is still 2.

patch/preconfig/packaging-update-abiname-to-11-2.patch
```diff
@@ -0,0 +1,20 @@
+Update debian abiname to 9-2
+
+From: Guohan Lu <gulv@microsoft.com>
+
+
```
qiluo-msft: Please also remove old packaging-update-abiname-to-9-2.patch, or rename it by git.

olivier-singla: Done

Makefile
olivier-singla: Done.
What about debian/config/defines, we still have abiname=11-2
The default abiname for debian 4.9.189 is 11, and we want a different abiname (such as 11-2)
If this is the same ABI, the script debian/bin/buildcheck.py will fail to verify the ABI

PR #122, opened 2020-02-07
Add ixgbe driver patch to support BCM54616S_E_PHY phy

jostar-yang:
This PR is patch for IXGBE.
0011-support-Broadcom-54616-Phy-for-Intel-igb-driver.patch is for IGB.
They are different code.
Do format with checkpatch.pl.

sonic-buildimage/src/sonic-linux-kernel/linux-4.9.168$ ./scripts/checkpatch.pl driver-ixgbe-external-phy.patch
total: 0 errors, 0 warnings, 0 checks, 37 lines checked

### DIFF
--- a/patch/driver-ixgbe-external-phy.patch
+++ b/patch/driver-ixgbe-external-phy.patch
@@ -1,0 +1,57 @@
+From: Jostar Yang <jostar_yang@accton.com>
+Signed-off-by: Jostar Yang <jostar_yang@accton.com>
+
+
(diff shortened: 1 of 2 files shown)


paulmenzel:
@jostar-yang, thank you for adding the support. According to the criteria in https://github.com/Azure/sonic-linux-kernel/blob/master/README.md, patches have to be sent upstream.

> Existing kernel modules need to be patched and those patches are available in upstream.

Have you done that already? Please also add the tag `Cc: stable@vger.kernel.org`, so it will be also added to the stable and LTS series.

PR #123, opened 2020-02-07
Fix i2c ISMT DMA buffer alignment issue

aravindmani-1:
Add linux patch to fix i2c ISMT 16-byte alignment issue of the DMA buffer address.


lguohan:
please identify the upstream commit id and the start linux version that introduce this patch.

lguohan:
use this for example. 

https://github.com/Azure/sonic-linux-kernel/blob/master/patch/linux-4.13-thermal-intel_pch_thermal-Fix-enable-check-on.patch

aravindmani-1:
Thanks @lguohan  for your comments.
Added the commit ID of the original patch.
Please let me know if this looks fine.

PR #129, opened 2020-04-03
[Mellanox] Add new patches come along with hw-mgmt V.7.0000.3020

keboliu:
Add new patches which required by new hw-mgmt package V.7.0000.3020
Including:

- support for VID protocol detection per page bases

- Increase i2c-mlxcpld transaction pooling

- mlxreg-hotplug: Use capability register for attribute creation

- Add support for attributes composed from few registers

- Add config for sensor xdpe122 and modify mlx_wdt to m


lguohan:
can you resolve conflict?

PR #130, opened 2020-04-18
Fix regulator issue on max6658 on buster

byu343:
CONFIG_REGULATOR is enabled by default on buster kernel.  With such change, devm_regulator_get() starts to return error code on the Arista 7050, which is also related to acpi=off being set on the platform. This fix is to ignore the error to keep max6658 initialized.

### DIFF
--- a/patch/driver-hwmon-max6658-bypass-regulator-error.patch
+++ b/patch/driver-hwmon-max6658-bypass-regulator-error.patch
@@ -1,0 +1,34 @@
+diff --git a/drivers/hwmon/lm90.c b/drivers/hwmon/lm90.c
+index 9d1a68d99..55c8913a4 100644
+--- a/drivers/hwmon/lm90.c
++++ b/drivers/hwmon/lm90.c
+@@ -1793,19 +1793,19 @@ static int lm90_probe(struct i2c_client *client,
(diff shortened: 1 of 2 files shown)


lguohan:
the patch ignores the devm_regulator_get error. it is ok to ignore on arista platform because acpi=off. can we ignore all such error on other platforms?

byu343:
> the patch ignores the devm_regulator_get error. it is ok to ignore on arista platform because acpi=off. can we ignore all such error on other platforms?

The regulator for the sensor here is unnecessary for Arista platforms, but we'd better not affect other platforms by making this change. 

We are testing an alternative solution of explicitly calling requlator_has_full_contraints() in arista platform driver.  If it is passed, this pull request can be discarded.

The reason behind it is that “acpi=off” makes difference to the error by not calling requlator_has_full_contraints(): https://elixir.bootlin.com/linux/latest/source/drivers/acpi/bus.c#L1109, 
which blocks the dummy regulator creation in devm_regulator_get().

lguohan:
is this patch still needed? if not, can we close this pr?

byu343:
It is not needed. We can close this pull request. The issue has been fixed in arista platform driver.

PR #131, opened 2020-04-18
Make amd iommu_disable safer

byu343:
This change is to fix a kernel issue that we found when using buster kernel on Arista 7050s with amd_iommu=off.


lguohan:
included in upstream v5.3

PR #134, opened 2020-05-03
[Patch] Armada i2c driver

antony-rheneus:
Enabled i2C MUX for GPIO
Safe protection lock for i2c transaction

Signed-off-by: Antony Rheneus <arheneus@marvell.com>


lguohan:
can you separate into two patch file? for the safe protection lock, is that upstream patch? does it pass the check-patch.pl test?

antony-rheneus:
> can you separate into two patch file? for the safe protection lock, is that upstream patch? does it pass the check-patch.pl test?

Ok, will move it two patches.
It needs to be reviewed by internal soc team before getting into to upstream. 
The base code itself without patch doesn't pass check-patch.pl.

PR #135, opened 2020-05-07
Add Backward porting the I2C ocores controller driver from kernel ups…

jjangmua:
Add backward porting the I2C ocores controller driver from kernel upstream version 5.1 to 4.19.

Containing upstream patches:
- e7663ef i2c: ocores: stop transfer on timeout
- 2dc9834 i2c: ocores: do not handle IRQ if IF is not set
- 69c8c0c i2c: ocores: add polling interface
- 2c7e492 i2c: ocores: add SPDX tag
- fac9c29 i2c: ocores: checkpatch fixes
- 809445d i2c: ocores: Add support for IO mapper registers
- 237b5f6 i2c: ocores: Add support for bus clock via platform data
- 088a8a7 i2c: ocores: turn incomplete kdoc into a comment

### DIFF
--- a/patch/driver-i2c-ocores-5-1.patch
+++ b/patch/driver-i2c-ocores-5-1.patch
@@ -1,0 +1,540 @@
+Add Backward porting the I2C ocores controller driver from kernel upstream version 5.1 to 4.19.
+
+Containing upstream patches:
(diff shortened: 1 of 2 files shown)


lguohan:
thanks!

PR #140, opened 2020-05-30
do not build DOCS and cloud-amd64

taocy001:
speed up kernel compile.
1. DO_DOCS=False do not skip generate document package on 4.19, use DEB_BUILD_PROFILES=nodoc instead.
2. use arch_amd64_none_amd64 to build the amd64 kernel package, avoid generating cloud-amd64 package.
    As a result, jenkins pipeline linux-kernel-buster-build-pr job `#31` takes 1h 6min, compare to `#30` takes 1h 24min.
https://sonic-jenkins.westus2.cloudapp.azure.com/view/all/job/common/job/linux-kernel-buster-build-pr/

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -75,7 +75,7 @@ $(addprefix $(DEST)/, $(MAIN_TARGET)): $(DEST)/% :
 
 	# generate linux build file for amd64_none_amd64
 	fakeroot make -f debian/rules.gen DEB_HOST_ARCH=armhf setup_armhf_none_armmp
-	fakeroot make -f debian/rules.gen DEB_HOST_ARCH=arm64 setup_arm64_none
+	fakeroot make -f debian/rules.gen DEB_HOST_ARCH=arm64 setup_arm64_none_arm64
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -93,11 +93,11 @@ $(addprefix $(DEST)/, $(MAIN_TARGET)): $(DEST)/% :
 	stg import -s ../patch/series
 
 	# Building a custom kernel from Debian kernel source
-	ARCH=$(CONFIGURED_ARCH) DEB_HOST_ARCH=$(CONFIGURED_ARCH) DO_DOCS=False fakeroot make -f debian/rules -j $(shell nproc) binary-indep
+	ARCH=$(CONFIGURED_ARCH) DEB_HOST_ARCH=$(CONFIGURED_ARCH) DEB_BUILD_PROFILES=nodoc fakeroot make -f debian/rules -j $(shell nproc) binary-indep
```
lguohan, line 96: i think we should apply nodoc for all three arch

PR #143, opened 2020-06-03
Kernel configuration changes

lguohan:
The Debian kernel used with SONiC includes almost all available hardware that
can be found on a system using Linux. This increases considerably the time
needed to build the kernel Debian image. Since there are many drivers,
protocols or filesystems which will never be used on a switch, there is a
imple mechanism to remove kernel options which does not require creating a
kernel configuration patch. The options which need to be excluded from the
kernel are simply listed in a flat text file:
patch/kconfig-exclusions

Example:

    CONFIG_REISERFS_FS
    CONFIG_JFS_FS
    CONFIG_XFS_FS

Similarly, there is a mechanism to include some kernel options by listing
these options into the flat text file:
patch/kconfig-inclusions

Example:

    CONFIG_INT340X_THERMAL=m
    CONFIG_LOG_BUF_SHIFT=18

If the files patch/kconfig-exclusions and patch/kconfig-inclusions exist,
they will be processed after all the kernel patches described in the patch
directory have been applied, exclusions being done before inclusions.

Also, the final kernel configuration will be checked to verify that:
- all options asked to be excluded are effectively not present in the kernel,
- and all options asked to be included are effectively present in the kernel,
using the exact type (module or buit-in) or string or number.


lguohan:
@olivier-singla to review

lguohan:
I compare the kconfig between this one and current master, they are the same. so, this pr is good.

PR #154, opened 2020-07-21
Issue #153 patch psample to safely unload the module

padmanarayana:
Address https://github.com/Azure/sonic-linux-kernel/issues/153 by removing the __ro_after_init macro.
Checked that with this patch, it is possible to remove the module safely.


patch/net-psample-module-unload.patch
```diff
@@ -0,0 +1,22 @@
+diff --git a/net/psample/psample.c b/net/psample/psample.c
```
lguohan, line 1: can you provide the description in this patch file?

padmanarayana: updated

PR #163, opened 2020-10-08
[arista] Reassign prefetch memory per platform

Nakrez:
This patch releases prefecth memory resources assigned by the BIOS and
reassigns them based on Arista platform specific values.

The patch has no effect on non Arista devices. In order for it to take
effect, there needs to be reassigne_prefmem=1 and
platform=AN_ARISTA_PLATFORM in the kernel command line.


lguohan:
can you sign CLA?

lguohan (approved):

paulmenzel:
Have you sent this upstream yet?

PR #164, opened 2020-10-21
Update optoe driver to add CMIS (QSFP-DD, OSFP, ...) 

mudsut4ke:
Porting back optoe driver update from : https://github.com/Azure/sonic-linux-kernel/pull/96

* Update to optoe driver to improve EOF handling

	Added patch: 'driver-support-optoe-EOF_fix.patch' to patch directory
	Added the patch to 'series

* Patch to add support for CMIS (QSFP-DD, OSFP, ...) class devices
to optoe driver.  Backward compatible, still supports SFP*, QSFP*


zhenggen-xu:
What is the impact without this patch?

mudsut4ke:
@zhenggen-xu , Without this patch we will not be able to read all the data contained in the CMIS device class EEPROM (QSFP-DD, OSFP, ...).

PR #168, opened 2020-11-05
[Mellanox] Add kernel patches with hw-mgmt V.7.0010.1300

keboliu:
Add new kernel patches for mlxsw driver:

-  Move PSU EEPROM configuration from kerenl to user space for SPC2 / SPC3 system 
   official  commit ID  c071afcea6ecf24a3c119f25ce9f71ffd55b5dc2 https://github.com/torvalds/linux/commit/c071afcea6ecf24a3c119f25ce9f71ffd55b5dc2


keboliu:
removed this redundant patch, it has already covered in 0020-mlxsw-core-Set-different-thermal-polling-time-based.patch
- mlxsw: core: thermal: Enable thermal zones by default

liat-grozovik:
@lguohan can you please approve? once done we will update submodule with prev fix as well.

PR #173, opened 2020-11-16
add internal patch marker

lguohan:
all internal patches will be add below the marker. This eases
internal patch management when public patches change

### DIFF
--- a/patch/series
+++ b/patch/series
@@ -77,7 +77,7 @@ driver-ixgbe-external-phy.patch
 0021-platform-x86-mlx-platform-Remove-PSU-EEPROM-configur.patch
 ############################################################
 #
-# Internal patches below (placeholder)
+# Internal patches will be added below (placeholder)
 # Do not add any public patch below this marker
 #
 ############################################################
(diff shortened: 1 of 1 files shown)


yxieca (approved):

paulmenzel:
Nit: will be add*ed*

I wondor if adding a directory `patch/internal` would be more beneficial, as then people just seeing the directory tree and not having looked at the file `patch/series` are able to categorize the patches.

lguohan:
@paulmenzel , for all sonic repos, we have a unified way to introduce additional patch on those repro. The patch will first those repros, and then build those repo. In the internal repo, we will then first patch this series file, and then use the patches series file to patch the kernel build.

paulmenzel:
> @paulmenzel, for all sonic repos, we have a unified way to introduce additional patch on those repro. The patch will first patch those repros, and then build those repo. In the internal repo, we will then first patch this series file, and then use the patches series file to patch the kernel build.

Sorry, I do not understand your comment. How does that contradict having a directory `patch/internal/…`?

lguohan:
let me try again. so, if we change the sonic-linux-kernel build process and introduce a new directory `patch/internal`, yes that works. however, such scheme is per-repo basis. for a sonic image, we have other repos which also needs internal patch, and those repros have different build process, having a `patch/internal` for those repo may not be the best way. 

Therefore, we have design a unified way to patch each submodule in `sonic-buildimage` repo. The assumption of the unified way is not to change build process in each submodule. In this case, we will patch `patch/series` in `sonic-linux-kernel` repo. 

The unified approach can be found here. https://github.com/Azure/sonic-buildimage/blob/master/slave.mk#L371

PR #177, opened 2020-12-01
[backport]: i2c: mux: pca954x: allow management of device idle state via sysfs

lguohan:
f1fb64b04b: driver-i2c-mux-pca954x-allow-management-of-device-idle-stat.patch (v5.2-rc1)
ddd7c492d: i2c: mux: pca954x: remove support for unused platform data (v5.2-rc1)

tested on celestica dx010

### DIFF
--- a/patch/driver-i2c-mux-pca954x-allow-management-of-device-idle-stat.patch
+++ b/patch/driver-i2c-mux-pca954x-allow-management-of-device-idle-stat.patch
@@ -1,0 +1,207 @@
+From f1fb64b04bf414ab04e31ac107bb28137105c5fd Mon Sep 17 00:00:00 2001
+From: Robert Shearman <robert.shearman@att.com>
+Date: Thu, 28 Feb 2019 11:43:43 +0000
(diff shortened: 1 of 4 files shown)


paulmenzel:
On what device is this tested?

It’d be great, if you sent a backport request to stable@vger.kernel.org, so it gets added to the stable series.

lguohan:
tested here. https://github.com/Azure/sonic-buildimage/pull/6429

paulmenzel:
Thank you for the reference. Commits should still be self-contained, and people cannot be expected to crossreference commits or even repositories. It’d be great, if the Celestica DX010 would be mentioned here too.

Also, I really think a message to stable@vger.kernel.org with the backport request should be sent. (And soon SONiC should switch to Linux 5.10 anyway. ;-))

lguohan:
added the tested message.  those patches will be removed 5.2-rc1. I haven't sent patch to stable@vger.kernel.org before. let me see what needs to be done before send to that mailing list.

PR #179, opened 2020-12-17
Fix manage-config option comparison issue

kuanyu99:
- What I did

To make **manage-config** able to compare string type kernel options.

- How I did it

Use **sed** to remove the double quote at prefix and suffix of the option value if it exists.

- How to verify it

Take `CONFIG_MODULE_SIG_KEY="debian/certs/test.pem"` in kconfig-inclusions for example.
Originally, if I give a kernel options with a double quoted string type value, it will return comparison failure.
```
Checking added kernel options...
Option CONFIG_MODULE_SIG_KEY should be set to ["debian/certs/test.pem"] instead of [debian/certs/test.pem]
```

After modification, it can pass the verification.

### DIFF
--- a/manage-config
+++ b/manage-config
@@ -108,6 +108,7 @@ if [ -e ${exclusion_file} -o -e ${inclusion_file} ]; then
             if [ ! -z "$opt" ] && [[ ! "$opt" =~ ^#.* ]]; then
                 n=${opt%=*}
                 v="${opt#*=}"
+                v=$(sed -e 's/^"//' -e 's/"$//' <<< "$v")
                 s=$(scripts/config --file ${CONFIG_FILE} -k --state $n)
                 if [ ! "$s" = "$v" ]; then
                     ret=2


lguohan:
are we expecting to have quote in the kconfig? why do you need to pass the quote?

kuanyu99:
> are we expecting to have quote in the kconfig? why do you need to pass the quote?

If there is no quote between the string, it will be come like this.
```
Checking added kernel options...
Option CONFIG_MODULE_SIG_KEY should be set to [debian/certs/test.pem] instead of []
```

The debian/build/build_amd64_none_amd64/.config can't correctly generated. It will become like `CONFIG_MODULE_SIG_KEY=""`.
Seems the slash inside the string mess up the script if no double quote it.

manage-config
```diff
@@ -108,6 +108,7 @@ if [ -e ${exclusion_file} -o -e ${inclusion_file} ]; then
             if [ ! -z "$opt" ] && [[ ! "$opt" =~ ^#.* ]]; then
                 n=${opt%=*}
                 v="${opt#*=}"
+                v=$(sed -e 's/^"//' -e 's/"$//' <<< "$v")
```
paulmenzel, line 111: As Bash’s [parameter expansion](https://manpages.debian.org/buster/bash/bash.1.en.html#Parameter_Expansion) is already used above, could you please also do this, instead of using *sed*?

paulmenzel:
If you added the example from the merge/pull request description to the commit message, that would be awesome.

manage-config
kuanyu99: Sure, done it.

PR #180, opened 2020-12-18
ip6 ra

praveen-li:
Test PR, Do not review, I will close later.

PR #182, opened 2021-01-05
[patch]: Allow user to set metric on default route learned via Router…

praveen-li:
… Advertisement.

Allow user to set metric on default route learned via Router Advertisement.
Note: RFC 4191 does not say anything for metric for IPv6 default route.

Fix:
For IPv4, default route is learned via DHCPv4 and user is allowed to change
metric using config in etc/network/interfaces. But for IPv6, default route can
be learned via RA, for which, currently a fixed metric value 1024 is used.

Ideally, user should be able to configure metric on default route for IPv6
similar to IPv4. This fix adds sysctl for the same.

Logs:
----------------------------------------------------------------
For IPv4:
----------------------------------------------------------------

Config in etc/network/interfaces
----------------------------------------------------------------
```
auto eth0
iface eth0 inet dhcp
    metric 4261413864
```

IPv4 Kernel Route Table:
----------------------------------------------------------------
```
$ sudo route -n
Kernel IP routing table
Destination     Gateway         Genmask         Flags Metric Ref    Use Iface
0.0.0.0         172.11.44.1     0.0.0.0         UG    -33553432 0        0 eth0
```

FRR Table, if default route is learned via routing protocol too.
----------------------------------------------------------------
```
Codes: K - kernel route, C - connected, S - static, R - RIP,
       O - OSPF, I - IS-IS, B - BGP, P - PIM, E - EIGRP, N - NHRP,
       T - Table, v - VNC, V - VNC-Direct, A - Babel, D - SHARP,
       > - selected route, * - FIB route

S>* 0.0.0.0/0 [20/0] is directly connected, eth0, 00:00:03
K   0.0.0.0/0 [254/1000] via 172.21.47.1, eth0, 6d08h51m
```

----------------------------------------------------------------
i.e. User can prefer Default Router learned via Routing Protocol,
Similar behavior is not possible for IPv6, without this fix.

----------------------------------------------------------------
After fix [for IPv6]:
----------------------------------------------------------------
```
sudo sysctl -w net.ipv6.conf.eth0.net.ipv6.conf.eth0.accept_ra_defrtr_metric=0x770003e9
```

IP monitor:
----------------------------------------------------------------
```
default via fe80::be16:65ff:feb3:ce8e dev eth0 proto ra metric 1996489705  pref high
```

Kernel IPv6 routing table
----------------------------------------------------------------
```
Destination                    Next Hop                   Flag Met Ref Use If
::/0                           fe80::be16:65ff:feb3:ce8e  UGDAe 1996489705 0
 0 eth0
```

FRR Routing Table, if default route is learned via routing protocol.
```
----------------------------------------------------------------
Codes: K - kernel route, C - connected, S - static, R - RIPng,
       O - OSPFv3, I - IS-IS, B - BGP, N - NHRP, T - Table,
       v - VNC, V - VNC-Direct, A - Babel, D - SHARP,
       > - selected route, * - FIB route

S>* ::/0 [20/0] is directly connected, eth0, 00:00:06
K   ::/0 [119/1001] via fe80::be16:65ff:feb3:ce8e, eth0, 6d07h43m
----------------------------------------------------------------
```

### DIFF
--- a/patch/kernel_v5.patch
+++ b/patch/kernel_v5.patch
@@ -1,0 +1,189 @@
+diff --git a/Documentation/networking/ip-sysctl.txt b/Documentation/networking/ip-sysctl.txt
+index 5b896a9..c80379e 100644
+--- a/Documentation/networking/ip-sysctl.txt
++++ b/Documentation/networking/ip-sysctl.txt
+@@ -1421,6 +1421,12 @@ accept_ra_defrtr - BOOLEAN
+ 	Functional default: enabled if accept_ra is enabled.
+ 			    disabled if accept_ra is disabled.
+
++accept_ra_defrtr_metric - INTEGER
++    Metric for default router learned in Router Advertisement.
++
++    Functional default: 0 if accept_ra_defrtr is enabled.
++                        -1 if accept_ra_defrtr is disabled.
++
+ accept_ra_from_local - BOOLEAN
+ 	Accept RA with source-address that is found on local machine
+         if the RA is otherwise proper and able to be accepted.
+diff --git a/include/linux/ipv6.h b/include/linux/ipv6.h
+index b9dfca5..6c9afff 100644
+--- a/include/linux/ipv6.h
++++ b/include/linux/ipv6.h
+@@ -30,6 +30,7 @@ struct ipv6_devconf {
+ 	__s32		max_desync_factor;
+ 	__s32		max_addresses;
+ 	__s32		accept_ra_defrtr;
++	__s32		accept_ra_defrtr_metric;
+ 	__s32		accept_ra_min_hop_limit;
+ 	__s32		accept_ra_pinfo;
+ 	__s32		ignore_routes_with_linkdown;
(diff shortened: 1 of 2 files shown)


paulmenzel:
Is that the same as merge/pull request https://github.com/Azure/sonic-linux-kernel/pull/180?

praveen-li:
> Is that the same as merge/pull request #180?

Yes it is, both fixes are almost similar except route structure is changed from rt6_info to fib6_info, so I have to adjust fix as below in one line:

Replace "rt->fib6_metric" with  "rt->rt6i_metric" for previous version.

lguohan:
why do we have two pr then? can you delete one?

praveen-li:
> why do we have two pr then? can you delete one?

@lguohan : I am not able to build the same patch for 4.9.168, So raised it, thinking, we have build tests for the 201811 branch as well. I will mark this as a Draft as well. Note: Both PRs are in the draft as of now.

PR #183, opened 2021-01-11
[dni_dps460] Fix read and write failure to ‘fan1_target’ attribute

ArunSaravananBalachandran:
To fix read and write failure to ‘fan1_target’ attribute of ‘dni_dps460’ driver.

- Updated the `pmbus_data` structure in ‘dni_dps460’ driver’s patch to the one present in ‘pmbus_core.c’ in the latest kernel version used.
- Reverted patch to use ‘kstrtol_from_user’ instead of ‘kstrtol’ for set operation in the driver.

Logs: [UT_logs.txt](https://github.com/Azure/sonic-linux-kernel/files/5795133/UT_logs.txt)

Resolves Azure/sonic-buildimage#6140.


paulmenzel:
Thank you for the patch.

1. Why can `patch/driver-hwmon-pmbus-dni_dps460-use-kstrtol-from-user.patch` be removed?
2. Could you please add the merge/pull request description also to the git commit message?
3. Please sign off the commit *and* `patch/driver-hwmon-pmbus-dni_dps460-update-pmbus-core.patch`.
4. Please send the patch to Linux upstream, so it does not need to be maintained in the future in here, but is maintained in upstream Linux.

Item 4 is very important in my opinion.

lguohan:
have the same question, @ArunSaravananBalachandran, can you address?

ArunSaravananBalachandran:
> Thank you for the patch.
> 
> 1. Why can `patch/driver-hwmon-pmbus-dni_dps460-use-kstrtol-from-user.patch` be removed?
> 2. Could you please add the merge/pull request description also to the git commit message?
> 3. Please sign off the commit _and_ `patch/driver-hwmon-pmbus-dni_dps460-update-pmbus-core.patch`.
> 4. Please send the patch to Linux upstream, so it does not need to be maintained in the future in here, but is maintained in upstream Linux.
> 
> Item 4 is very important in my opinion.


1. With `driver-hwmon-pmbus-dni_dps460-use-kstrtol-from-user.patch` introduced to avoid the crash seen in #152, write on the attribute fails with 'Bad address' error. 
```
root@sonic:/home/admin# echo 15000 > /sys/class/i2c-adapter/i2c-1/1-0059/fan1_target
bash: echo: write error: Bad address
root@sonic:/home/admin#
```
Updating the pmbus_data structure in ‘dni_dps460’ driver’s patch to the one used in latest kernel version, resolves the crash seen in both read Azure/sonic-buildimage#6140 and write #152 on the attribute and removing the `driver-hwmon-pmbus-dni_dps460-use-kstrtol-from-user.patch` resolves the 'Bad address' error.

2. Will amend the commit to update the commit message.
3. Will signoff the commit and the patch through a new commit.
4. The `dni_dps460` driver is defined only in sonic-linux-kernel via `patch/driver-hwmon-pmbus-dni_dps460.patch` and not available in upstream linux. Currently this driver is used only in Dell S6000 platform for accessing PSU attributes.

paulmenzel:
Thank you for your replies. Your explanation in item 1. is very good, and is perfect to have in a commit message, so people do not have to read the merge/pull request discussion.

Regarding 4., I know the driver is not in upstream Linux, but that is the goal, and nothing would prevent that. To avoid future porting work, it’d be great, if you sent the whole driver upstream, seeing that you also have access to the hardware.

lguohan (approved):

PR #186, opened 2021-01-14
[201811]: security update kernel to 4.9.246

lguohan:

### DIFF
--- a/patch/preconfig/series
+++ b/patch/preconfig/series
@@ -1,1 +1,1 @@
-changelog.patch
+# changelog.patch
--- a/Makefile
+++ b/Makefile
@@ -2,10 +2,10 @@
 SHELL = /bin/bash
 .SHELLFLAGS += -e
 
-KVERSION_SHORT ?= 4.9.0-9
+KVERSION_SHORT ?= 4.9.0-14
 KVERSION ?= $(KVERSION_SHORT)-amd64
-KERNEL_VERSION ?= 4.9.168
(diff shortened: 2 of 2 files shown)


patch/preconfig/series
```diff
@@ -1 +1 @@
-changelog.patch
+# changelog.patch
```
qiluo-msft, line 1: Will you remove this line and this file?

qiluo-msft (commented):
As comment

patch/preconfig/series
lguohan: not necessary

qiluo-msft (approved):

paulmenzel:
Thank you for the update.

1.  Please split this into three commits with respective commit messages.
    a)  patch/series: Remove changelog.patch
    b)  Introduce variable for base URL
    c)  Update kernel to 4.9.246 (the tag is not needed in the commit message summary, as it’s in the correct branch)
2.  Also, it’d be great if you mentioned the security issue being fixed in the commit message of the commit updating the Linux version.
3.  Lastly, please mention if this has been tested somewhere.

lguohan:
b) and c) is hard to split since base_url also changed.

added changelog for security issues from debian upstream

since it is taken from debian kernel, so we just did the build test.

paulmenzel:
> b) and c) is hard to split since base_url also changed.

Next time, first introduce the variable with the “old” URL, and then change the URL.

PR #193, opened 2021-01-27
[patch]: optoe update to test changes for upstream version

donboll:
I am preparing optoe for submittal to the upstream kernel (5.11).  These changes have been built against several kernel versions, and tested on several platforms.  I am requesting that this pull request be built and tested in SONiC before the upstream submittal.

This is not the production version!  There may be changes resulting from the upstream review, and at the least this version will need '#DEFINE DEBUG 1' removed.


prgeor:
@donboll why is this change required? Current optoe won't work with 5.11 kernel?

donboll:
This request was submitted 2+ years ago.  The request was to have the Sonic team test proposed changes to the optoe driver prior to submitting that driver upstream.  Sonic is the most robust user and tester of optoe so I thought it would be productive to test BEFORE submitting upstream.

The upstream folks rejected optoe outright, they don't want any code that access network components (eg optical devices) that doesn't work through the kernel network framework.  They had no problem with the correctness of the code, they just did not want that capability in the upstream kernel.

This request is moot.  There is no need for the test since the upstream submittal was rejected.

donboll:
This request was submitted 2+ years ago.  The request was to have the Sonic team test proposed changes to the optoe driver prior to submitting that driver upstream.  Sonic is the most robust user and tester of optoe so I thought it would be productive to test BEFORE submitting upstream.

The upstream folks rejected optoe outright, they don't want any code that access network components (eg optical devices) that doesn't work through the kernel network framework.  They had no problem with the correctness of the code, they just did not want that capability in the upstream kernel.

This request is moot.  There is no need for the test since the upstream submittal was rejected.

PR #198, opened 2021-03-02
[net] Exclude net prio and classid cgroups to avoid conflict with cgroup2

maksymbelei95:
* Unset CONFIG_CGROUP_NET_CLASSID, CONFIG_CGROUP_NET_PRIO and all depended
  modules in Kconfig to disable those v1 cgroups. This will resolve conflict
  with cgroup2, thus, programs which use cgroup2 will work normally.

Signed-off-by: Maksym Belei <Maksym_Belei@jabil.com>


paulmenzel:
Thank you for the patch. Can you please add the details, how to reproduce this? I would have imagined, if there was a conflict, than the Linux would have measure to prevent it.

maksymbelei95:
@paulmenzel, could you check the related defect: https://github.com/Azure/sonic-buildimage/issues/6858? There is information about how to reproduce the issue and my investigation result with evidences regarding conflict of usage of cgroups v1 and v2. Also, here is 3rd party source with information, regarding net_prio, net_cls and socket matching in cgroup2.

I am currently testing the change locally, could somebody check it with pure jenkins build?

paulmenzel:
Thank you for your quick reply. A commit message should be self-contained in my opinion. I am going to check in the next days, but it’d be great, if you extended the commit message.

maksymbelei95:
Commit message has updated, could you check it?

paulmenzel (approved):

maksymbelei95:
@lguohan, the issue is being occurred on 202012 too. I think we can cherry-pick the solution to 202012 without any changes. Could you mark it as for 202012?

dprital:
@daall - Can you please merge to 202012 ?

PR #208, opened 2021-04-06
[max6620] Use Fan Dynamics registers for speed calculation and add attributes to retrieve fan fault status

ArunSaravananBalachandran:
Incorporate following enhancements in MAX6620 driver:

- Use value of Fan Dynamics registers for tach count to speed (in RPM) conversion and vice versa
- Add sysfs attributes (fan*_alarm) to retrieve fan fault status

Signed-off-by: Arun Saravanan Balachandran <Arun_Saravanan_Balac@dell.com>

### DIFF
--- a/patch/driver-hwmon-max6620-add-alarm-attribute.patch
+++ b/patch/driver-hwmon-max6620-add-alarm-attribute.patch
@@ -1,0 +1,66 @@
+From f84450af3a86368a64c36c114a2a69068a130d47 Mon Sep 17 00:00:00 2001
+From: Arun Saravanan Balachandran <Arun_Saravanan_Balac@dell.com>
+Date: Wed, 17 Mar 2021 18:04:29 +0530
+Subject: [PATCH] Add attributes in MAX6620 driver to retrieve fan fault status
+
+Signed-off-by: Arun Saravanan Balachandran <Arun_Saravanan_Balac@dell.com>
+---
+ drivers/hwmon/max6620.c | 18 +++++++++++++-----
+ 1 file changed, 13 insertions(+), 5 deletions(-)
+
+diff --git a/drivers/hwmon/max6620.c b/drivers/hwmon/max6620.c
+index 5c8b155..6f921bb 100644
+--- a/drivers/hwmon/max6620.c
++++ b/drivers/hwmon/max6620.c
+@@ -283,6 +283,9 @@ static ssize_t set_target(struct device *dev, struct device_attribute *devattr,
+ 	i2c_smbus_write_byte_data(client, target_reg[attr->index], target1);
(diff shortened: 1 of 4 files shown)


lguohan:
upstream!

ArunSaravananBalachandran:
> upstream!

@lguohan, please note that the current max6620 has several earlier patches applied and this one builds on top of those earlier patches. w.r.t upstreaming, should I consolidate and upstream all these patches to [lgrunenberg/max6620](https://github.com/lgrunenberg/max6620) ?

jleveque:
@paulmenzel: Could you please provide us with some guidance on how we could upstream this max6620 code? Basically, in SONiC, we are currently using the code from lgrunenberg's repo and applying our own patches on top of it. It would be great if we could upstream it all.

lguohan:
> Use value of Fan Dynamics registers for tach count to speed (in RPM) conversion and vice versa

what is the issue it tries to solve?

paulmenzel:
> @paulmenzel: Could you please provide us with some guidance on how we could upstream this max6620 code? Basically, in SONiC, we are currently using the code from lgrunenberg's repo and applying our own patches on top of it. It would be great if we could upstream it all.

Thank you for picking this up. I would create a new branch in my Linux git folder based on current git://git.kernel.org/pub/scm/linux/kernel/git/groeck/linux-staging.git, then add the patches one-by-one with the author information from the original authors, after making sure they build, add my Signed-off-by line, and run `git format-patch -X out`, where X is the number of commits, and then `git send-email out/*`.

ArunSaravananBalachandran:
> > Use value of Fan Dynamics registers for tach count to speed (in RPM) conversion and vice versa
> 
> what is the issue it tries to solve? what is fan dynamics registers? why do we use it now?

Fan Dynamics can be used to configure the nominal RPM range of the fan. The ‘Tach period counts’ configured in the register should be used for the tach count to speed in RPM conversion and vice versa.

By default the driver sets the register value to ‘2 Tach period counts’ and it results in approximation of target speeds for certain larger speed values.
The register value can be increased accordingly to reduce approximations in the higher speed ranges.  

In the current max6620 driver ‘fan*_div’ sysfs attribute is provided to configure the Fan dynamics register
But in spite of the value provided to ‘fan*_div’ being configured in the Fan dynamics register, the driver uses a hardcoded value of ‘2’ for the speed calculation resulting in incorrect values of speed set and read afterwards.

As part of this PR, the Fan dynamics register’s value is used for the speed calculation instead of a hardcoded value to resolve the above issue.

ArunSaravananBalachandran:
Closing this PR, as the driver changes in upstream have been backported to version 5.10 by #241 and 4.19 by #243

PR #221, opened 2021-06-29
Enable MDIO related kernel config options

msiddare:
By default, MDIO_BUS_MUX is set indirectly via other configurations (in
particular OF must be enabled).  However, we don't need OF, but still
wanted MDIO_BUS_MUX to be configured for accessing required MDIO drivers.

The patch here is to remove the dependence on OF_MDIO, and add a prompt/label
to MDIO_BUS_MUX so that it can be explicitly configured.

### DIFF
--- a/patch/preconfig/cisco-mdio.patch
+++ b/patch/preconfig/cisco-mdio.patch
@@ -1,0 +1,40 @@
+From 40effb100908437fa43fe0b8fa9c5802ea1cf162 Mon Sep 17 00:00:00 2001
+From: Madhava Reddy Siddareddygari <msiddare@cisco.com>
+Date: Tue, 9 Mar 2021 08:41:15 -0800
+Subject: [PATCH] Enable MDIO related options
+
+---
+ debian/config/amd64/config | 4 ++++
+ drivers/net/phy/Kconfig    | 4 ++--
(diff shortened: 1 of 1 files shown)


patch/preconfig/cisco-mdio.patch
```diff
@@ -0,0 +1,40 @@
...
+--- a/debian/config/amd64/config
++++ b/debian/config/amd64/config
+@@ -247,3 +247,7 @@ CONFIG_ZONE_DEVICE=y
+ ##
+ CONFIG_LSM_MMAP_MIN_ADDR=65536
+ 
++#
++CONFIG_PHYLIB=m
++CONFIG_OF_MDIO=m
++CONFIG_MDIO_BUS_MUX=m
```
lguohan, line 22: please use kconfig-inclusion for all kconfig changes.

msiddare: Hi Guohan,

There are dependencies with other kernel configs and some of the configs are enabled by selecting other configs which are not related to cisco configs.

This PR will make sure only minimum required configs are enabled with this PR

msiddare: Make changes in kconfig-inclusion and add any dependent configurations to kconfig-inclusion

PR #223, opened 2021-06-30
Enable REGMAP_I2C, REGMAP_MMIO kernel config options

msiddare:
REGMAP_I2C and REGMAP_MMIO would be indirectly configured by enabling
some other drivers, which cisco platform doesn't depend on. This patch
gives each a prompt/label so that we can explicitly configure them.
It would be possible to remove this patch by configuring some other dummy
driver(s) so as to ensure these two configs get enabled. However it would
lead to enabling extra unwanted kernel configs.

### DIFF
--- a/patch/preconfig/cisco-regmap.patch
+++ b/patch/preconfig/cisco-regmap.patch
@@ -1,0 +1,46 @@
+From 94efa7350c1630ca20ab2adabc7ef7452b8d4dc5 Mon Sep 17 00:00:00 2001
+From: Madhava Reddy Siddareddygari <msiddare@cisco.com>
+Date: Tue, 9 Mar 2021 09:49:12 -0800
+Subject: [PATCH] Enable REGMAP_I2C, REGMAP_MMIO
+
+---
+ debian/config/amd64/config  | 3 +++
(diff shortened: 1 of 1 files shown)


patch/preconfig/cisco-regmap.patch
```diff
@@ -0,0 +1,46 @@
...
+ #
+ CONFIG_ARCH_NR_GPIO=10240
++#
++CONFIG_REGMAP_I2C=m
++CONFIG_REGMAP_MMIO=m
+diff --git a/drivers/base/regmap/Kconfig b/drivers/base/regmap/Kconfig
+index 8cd2ac650..409cd726f 100644
+--- a/drivers/base/regmap/Kconfig
++++ b/drivers/base/regmap/Kconfig
+@@ -17,7 +17,7 @@ config REGMAP_AC97
```
lguohan, line 26: please add comments why you need add option. needed by external mfd driver.

REGMAP_I2C and REGMAP_MMIO would be indirectly configured by enabling
some other drivers, which cisco platform doesn't depend on. This patch
gives each a prompt/label so that we can explicitly configure them.

PR #226, opened 2021-07-01
Allow NULL of_node device

msiddare:

(cisco using ACPI for kernel module configuration).  The kernel
(fw_node) in the device structure.  It is unclear if they can be
using ACPI and not device tree.  In this particular driver, there
reference to the device tree was made within the driver.  This caused


patch/cisco-ds4424-null-of-node.patch
```diff
@@ -0,0 +1,23 @@
+commit 166b37e58be17655da39f3d7ad61014db4afa4a1
+Author: Billie R Alsup <balsup@cisco.com>
+Date:   Sun Aug 23 14:57:55 2020 -0700
+
+    Allow NULL of_node
+
```
paulmenzel, line 6: Please use `git format-patch`.

msiddare: As per the review comments patch updated to use git format-patch

patch/cisco-ds4424-null-of-node.patch
```diff
@@ -0,0 +1,45 @@
...
++++ b/drivers/iio/dac/ds4424.c
+@@ -236,12 +236,6 @@ static int ds4424_probe(struct i2c_client *client,
+ 	indio_dev->dev.of_node = client->dev.of_node;
+ 	indio_dev->dev.parent = &client->dev;
+ 
+-	if (!client->dev.of_node) {
+-		dev_err(&client->dev,
+-				"Not found DT.\n");
+-		return -ENODEV;
+-	}
```
lguohan, line 38: this one should be upstreamed by cisco per discussion.

paulmenzel:
Please take Linux commit a2d2010d95, and contact stable@vger.kernel.org to backport it to to 4.19.

```
$ git show a2d2010d95
commit a2d2010d95cd7ffe3773aba6eaee35d54e332c25
Author: Ruslan Babayev <ruslan@babayev.com>
Date:   Sun May 5 12:24:36 2019 -0700

    iio: dac: ds4422/ds4424 drop of_node check
    
    The driver doesn't actually rely on any DT properties. Removing this
    check makes it usable on ACPI based platforms.
    
    Signed-off-by: Ruslan Babayev <ruslan@babayev.com>
    Cc: xe-linux-external@cisco.com
    Signed-off-by: Jonathan Cameron <Jonathan.Cameron@huawei.com>

diff --git a/drivers/iio/dac/ds4424.c b/drivers/iio/dac/ds4424.c
index 883a47562055..2b3ba1a66fe8 100644
--- a/drivers/iio/dac/ds4424.c
+++ b/drivers/iio/dac/ds4424.c
@@ -236,12 +236,6 @@ static int ds4424_probe(struct i2c_client *client,
        indio_dev->dev.of_node = client->dev.of_node;
        indio_dev->dev.parent = &client->dev;
 
-       if (!client->dev.of_node) {
-               dev_err(&client->dev,
-                               "Not found DT.\n");
-               return -ENODEV;
-       }
-
        data->vcc_reg = devm_regulator_get(&client->dev, "vcc");
        if (IS_ERR(data->vcc_reg)) {
                dev_err(&client->dev,
```

lguohan (approved):

PR #227, opened 2021-07-12
Upgrade kernel from 4.19.152 to 5.10.46

saiarcot895:
As part of the Bullseye upgrade, upgrade to the 5.10.46 kernel. At this
point, all but one patch have been either ported from 4.19 or dropped
entirely.

Also update our build script to commit changes to *.maintscript and
*.bug-presubj files, these are generated files and get changed because
of our abiname change.


paulmenzel:
Tested how?

saiarcot895:
Not yet tested. This is still a WIP, and only a subset of patches have been ported to the 5.10.40 kernel (hence the draft status). More work still needs to be done.

lguohan:
looks like there is compilation issue, can you modify the azp so that we can compile it?

saiarcot895:
I've updated the pipeline file, but the image hasn't been published yet, so CI will still fail until that is resolved.

paulmenzel (commented):
Did you report these regressions to Linux upstream?

saiarcot895:
@paulmenzel I've started a mail thread [here](https://lore.kernel.org/netdev/BL0PR2101MB13167E456C5E839426BCDBE6D9CB9@BL0PR2101MB1316.namprd21.prod.outlook.com/T/#u).

paulmenzel:
Why upgrade from 5.10.40 to 5.10.46, and not the current 5.10.73?

saiarcot895:
5.10.46 was the kernel version at the time of release of Bullseye, and so I settled at that version. Time permitting, I can take a look at upgrading it to 5.10.73 (or whatever the latest version is at that time).

PR #228, opened 2021-07-13
[Marvell] CPU1 failure on continuous reboot

rajkumar38:
On continuous reboot, once in a while CPU1 (secondary cpu) doesn't come online. Issue seen on migrating to Debian gcc 8.3.6 from gcc7.
Until introducing the Debian GCC 8.x version – the execution process didn’t influence
mmu intialization before starting the socondary processor.
The new GCC version makes some optimizations that execute the SW faster and once in a while,
causing the Secondary CORE to start while the MMU is still in the initialization stage.
Once adding a slight delay right after the MMU initialization –
compensate the optimized code and enables the MMU to init successfully before the 2nd CORE starts.

Signed-off-by: Rajkumar Pennadam Ramamoorthy <rpennadamram@marvell.com>

### DIFF
--- a/patch/armhf_secondary_boot_online.patch
+++ b/patch/armhf_secondary_boot_online.patch
@@ -1,0 +1,56 @@
+From 500d48948f43d18a07894b5fd4f891afbade1d79 Mon Sep 17 00:00:00 2001
+From: Antony Rheneus <arheneus@marvell.com>
+Date: Thu, 1 Jul 2021 20:34:01 +0530
+Subject: [PATCH 1/1] From 28c199cfda1746644211893b6614acbac970515b Mon Sep 17
+ 00:00:00 2001 Subject: [PATCH 1/1] CPU1 failed to come online
+MIME-Version: 1.0
+Content-Type: text/plain; charset=UTF-8
+Content-Transfer-Encoding: 8bit
+
+Issue: On contiuous reboot once in a while CPU1 (secondary cpu) doesn't
+come online
+The issue comes after migrating to Debian gcc 8.3.6, the issue was not seen till gcc7
+
+Until introducing the Debian GCC 8.x version – the execution process didn’t influence 
+mmu intialization before starting the socondary processor.
+The new GCC version makes some optimizations that execute the SW faster and once in a while,
+causing the Secondary CORE to start while the MMU is still in the initialization stage.
+Once adding a slight delay right after the MMU initialization –
+compensate the optimized code and enables the MMU to init successfully before the 2nd CORE starts.
+
+Signed-off-by: Antony Rheneus <arheneus@marvell.com>
+---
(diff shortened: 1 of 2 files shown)


paulmenzel:
continuous in the title.

patch/armhf_secondary_boot_online.patch
```diff
@@ -0,0 +1,56 @@
+From 500d48948f43d18a07894b5fd4f891afbade1d79 Mon Sep 17 00:00:00 2001
+From: Antony Rheneus <arheneus@marvell.com>
+Date: Thu, 1 Jul 2021 20:34:01 +0530
+Subject: [PATCH 1/1] From 28c199cfda1746644211893b6614acbac970515b Mon Sep 17
```
paulmenzel, line 4: Is that commit in upstream Linux master? I didn’t see it.

patch/armhf_secondary_boot_online.patch
```diff
@@ -0,0 +1,56 @@
...
+ 00:00:00 2001 Subject: [PATCH 1/1] CPU1 failed to come online
```
paulmenzel, line 5: Two subjects look strange.

patch/armhf_secondary_boot_online.patch
```diff
@@ -0,0 +1,56 @@
...
+ extern void mvebu_cortex_a9_secondary_startup(void);
++static DECLARE_COMPLETION(cpu_wait);
+ 
+ static int mvebu_cortex_a9_boot_secondary(unsigned int cpu,
+ 						    struct task_struct *idle)
+@@ -44,6 +47,7 @@ static int mvebu_cortex_a9_boot_secondary(unsigned int cpu,
+ 		mvebu_pmsu_set_cpu_boot_addr(hw_cpu, mvebu_cortex_a9_secondary_startup);
+ 	smp_wmb();
+ 
++	wait_for_completion_timeout(&cpu_wait, usecs_to_jiffies(10));
```
paulmenzel, line 50: That looks like a hack. It’s 10 μs, right? Isn’t there a better way to ensure the MMU is initialized?

paulmenzel (commented):
On what exact model did you test this?

patch/armhf_secondary_boot_online.patch
lguohan: they are going to upstream it. the model is armhf.

lguohan: armada family.

patch/armhf_secondary_boot_online.patch
lguohan: later cpu has a register to query the mmu init status, but for this model it does not have. so, looks like the best they can do. When they upstream to linux kernel, they will have to explain everything clearly, then we will replace the patch with the upstreamed version.

PR #230, opened 2021-08-04
Increase optoe's write-max to support i2c block write

prgeor:
In current optoe driver, the value of write_max is hardcoded to one byte.

CMIS spec supports firmware upgrade on QSFP-DD transceivers which means there is significant overhead due to this one byte write limit when a platform wants to upgrade the firmware over an i2c bus at 100Khz.

The overhead is more pronounced when a platform has shared master and platform needs to perform firmware upgrade on multiple transceivers

Signed-off-by: Prince George <prgeor@microsoft.com>

### DIFF
--- a/patch/driver-support-optoe-write-max.patch
+++ b/patch/driver-support-optoe-write-max.patch
@@ -1,0 +1,22 @@
+diff --git a/drivers/misc/eeprom/optoe.c b/drivers/misc/eeprom/optoe.c
+index 16287fdc5..61c408a84 100644
+--- a/drivers/misc/eeprom/optoe.c
++++ b/drivers/misc/eeprom/optoe.c
+@@ -1069,7 +1069,7 @@ static int optoe_probe(struct i2c_client *client,
+ 		 * 2 byte writes are acceptable for PE and Vout changes per
+ 		 * Application Note AN-2071.
+ 		 */
+-		unsigned int write_max = 1;
++		unsigned int write_max = 32;
+ 
+ 		optoe->bin.write = optoe_bin_write;
+ 		optoe->bin.attr.mode |= 0200;
+@@ -1131,6 +1131,8 @@ static int optoe_probe(struct i2c_client *client,
+ 	dev_info(&client->dev, "%zu byte %s EEPROM, %s\n",
+ 		optoe->bin.size, client->name,
+ 		optoe->bin.write ? "read/write" : "read-only");
++    printk(KERN_ERR "$$$optoe_probe(): %zu byte %s EEPROM, write_max=%u\n",
++            optoe->bin.size, client->name, optoe->write_max);
+ 
+ 	if (use_smbus == I2C_SMBUS_WORD_DATA ||
+ 	    use_smbus == I2C_SMBUS_BYTE_DATA) {
--- a/patch/driver-support-optoe-write-max.patch
+++ b/patch/driver-support-optoe-write-max.patch
@@ -1,46 +1,86 @@
-From 83573e4c3171b528a4516caf3483cfccf90b9a32 Mon Sep 17 00:00:00 2001
+From ca9daf437f0910493a6eb8ced900662223ea3cf7 Mon Sep 17 00:00:00 2001
 From: Prince George <prgeor@microsoft.com>
-Date: Wed, 4 Aug 2021 10:06:18 +0000
-Subject: [PATCH] Increase optoe's write-max to support i2c block write
+Date: Mon, 16 Aug 2021 16:04:15 +0000
+Subject: [PATCH] Dynamic write_max support for optoe driver  
+ In current optoe driver, the value of write_max is  hardcoded to one byte.  
+ CMIS spec supports firmware upgrade on QSFP-DD transceivers which means  
+ there is significant overhead due to this one byte write limit when a
+ platform wants to upgrade the firmware over an i2c bus at 100Khz. 
+ The  overhead is more pronounced when a platform has shared master and 
+ platform needs to perform firmware upgrade  on multiple transceivers.
 
-In current optoe driver, the value of write_max is hardcoded to one byte.
-CMIS spec supports firmware upgrade on QSFP-DD transceivers which means
-there is significant overhead due to this one byte write limit when a
-platform wants to upgrade the firmware over an i2c bus at 100Khz. The
-overhead is more pronounced when a platform has shared master and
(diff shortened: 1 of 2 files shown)


patch/driver-support-optoe-write-max.patch
```diff
@@ -0,0 +1,22 @@
+diff --git a/drivers/misc/eeprom/optoe.c b/drivers/misc/eeprom/optoe.c
```
paulmenzel, line 1: Please add a `git format-patch` formatted patch with a commit message and Signed-off-by line.

prgeor:
I am keeping this as draft for testing. A proper fix is need by reading the write-max from the device tree node.

paulmenzel (commented):
Please add hard measured numbers to the commit message.

patch/driver-support-optoe-write-max.patch
prgeor: Added

patch/driver-support-optoe-write-max.patch
```diff
@@ -0,0 +1,22 @@
...
++		unsigned int write_max = 32;
+ 
+ 		optoe->bin.write = optoe_bin_write;
+ 		optoe->bin.attr.mode |= 0200;
+@@ -1131,6 +1131,8 @@ static int optoe_probe(struct i2c_client *client,
+ 	dev_info(&client->dev, "%zu byte %s EEPROM, %s\n",
+ 		optoe->bin.size, client->name,
+ 		optoe->bin.write ? "read/write" : "read-only");
++    printk(KERN_ERR "$$$optoe_probe(): %zu byte %s EEPROM, write_max=%u\n",
++            optoe->bin.size, client->name, optoe->write_max);
```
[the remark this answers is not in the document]
prgeor: followed in latest patch

patch/driver-support-optoe-write-max.patch
```diff
@@ -0,0 +1,46 @@
...
+diff --git a/drivers/misc/eeprom/optoe.c b/drivers/misc/eeprom/optoe.c
+index 16287fdc5..6cf61e645 100644
+--- a/drivers/misc/eeprom/optoe.c
++++ b/drivers/misc/eeprom/optoe.c
+@@ -1069,7 +1069,7 @@ static int optoe_probe(struct i2c_client *client,
+ 		 * 2 byte writes are acceptable for PE and Vout changes per
+ 		 * Application Note AN-2071.
+ 		 */
+-		unsigned int write_max = 1;
++		unsigned int write_max = 32;
```
Staphylo, line 27: So the decision here is to hardcode the `write_max` size to SMBus 2.x max block size?
I am mostly fine with this solution but the lack of configuration knob might lead to some xcvrs not being supported by optoe and might therefore make some users unhappy.

I think one of 2 mechanism could be used to allow users with known bad xcvrs to still have them working properly.

1) Adding a module parameter that can be tweaked at runtime and `write_max` would use this value.
This would offer a global knob for the module, but changing the `write_max` speed would impact all xcvrs even though only one might need this workaround.

2) Adding a per xcvr sysfs entry called `max_write_size` similarily to the `eeprom` and `port_name` entries that already exists.

In both cases above the choice of the default value should be chosen
 - default value of 32 because there is no reason to penalize the performance of xcvr models that follow the spec
 - default value of 1 for backward compatibility and likely enhanced reliability
In both these cases it still means that some logic needs to run somewhere (xcvrd) to change the `write_max` value based on what is seen.
AFAIK write speed doesn't matter that much on xcvrs for anything else than firmware upgrades, so potentially the logic performing the upgrade could upgrade write_size to 32.

But IMO the burden of adding quirks should be on crappy devices that don't implement the spec correctly.
I don't see why everyone else should suffer from the shortcomings of a brand/model.

prgeor: Thanks Samuel for reviewing. I will wait for the performance numbers  before making the write-max configurable.

prgeor: Fixed in latest commit

patch/driver-support-optoe-write-max.patch
```diff
@@ -0,0 +1,86 @@
...
+ 
++static ssize_t show_dev_write_max_size(struct device *dev,
++			struct device_attribute *dattr, char *buf)
++{
++	struct i2c_client *client = to_i2c_client(dev);
++	struct optoe_data *optoe = i2c_get_clientdata(client);
++	ssize_t count;
++
++	mutex_lock(&optoe->lock);
++	count = sprintf(buf, "%u\n", optoe->write_max);
```
Staphylo, line 40: I believe you want to use `scnprintf` here instead of `sprintf`.
Even though formatting a unsigned integer is unlikely to reach the `PAGE_SIZE` limit of `buf` it has the advantage of returning the size of the written buffer minus the `\0` char.
I believe this is the best practice since it's part of this documentation https://www.kernel.org/doc/html/latest/filesystems/sysfs.html#reading-writing-attribute-data

Staphylo (commented):
LGTM just one small comment

patch/driver-support-optoe-write-max.patch
prgeor: The (*show) API is not very particular about excluding the null char from the returned count. The preference of scnprintf() over sprintf() is more from overflow protection mentioned here:-  
[https://elixir.bootlin.com/linux/v5.9.16/source/Documentation/filesystems/sysfs.rst#L247]

I think we are pretty much safe here to use sprintf().

prgeor: [https://elixir.bootlin.com/linux/v5.9.16/source/Documentation/filesystems/sysfs.rst#L247](https://elixir.bootlin.com/linux/v5.9.16/source/Documentation/filesystems/sysfs.rst#L247)

Staphylo (approved):

patch/driver-support-optoe-write-max.patch
```diff
@@ -0,0 +1,86 @@
...
++	return count;
++}
++
++static ssize_t set_dev_write_max_size(struct device *dev,
++			struct device_attribute *attr,
++			const char *buf, size_t count)
++{
++	struct i2c_client *client = to_i2c_client(dev);
++	struct optoe_data *optoe = i2c_get_clientdata(client);
++	int write_max_size;
```
prgeor, line 52: one issue, write_max_size should be unsigned int to allow size > 127 bytes

lguohan (approved):

PR #224, opened 2021-06-30
Reserve address space for powered-off devices behind bridges

msiddare:
Reserve PCIe address space for powerd-off devices behind PCIe bridge.

Data path devices are powered off by default, they will not be visible
at BIOS stage and memory for these devices is not reserved.

By default, no address space would be reserved on the bridges for these
unpowered devices.  When they were powered up, they could fail to initialize
because there was no appropriately aligned window available for a given BAR.

This patch will reserve address space for data path devices that are
behind PCIe bridge, so that when devices are available PCIe subsystem
will be assign the address within the specified range.

Signed-off-by: Madhava Reddy Siddareddygari <msiddare@cisco.com>

PR #235, opened 2021-09-21
Reserve PCIe address space for powerd-off devices behind PCIe bridge.

msiddare:
Data path devices are powered off by default, they will not be visible
at BIOS stage and memory for these devices is not reserved.

By default, no address space would be reserved on the bridges for these
unpowered devices. When they were powered up, they could fail to initialize
because there was no appropriately aligned window available for a given BAR.

This patch will reserve address space for data path devices that are
behind PCIe bridge, so that when devices are available PCIe subsystem
will be assign the address within the specified range.


msiddare:
Have had issues submitting changes to pull https://github.com/Azure/sonic-linux-kernel/pull/224
Created a new pull request and incorporated changes in this PR. Will close other one.

lguohan (approved):

PR #236, opened 2021-09-26
Cisco/ds4424

msiddare:

(cisco using ACPI for kernel module configuration). The kernel
(fw_node) in the device structure. It is unclear if they can be
using ACPI and not device tree. In this particular driver, there
reference to the device tree was made within the driver. This caused


msiddare:
Changes in this pull request are reviewed in PR https://github.com/Azure/sonic-linux-kernel/pull/226
As I was not able to update PR226 due to permission issue, opened a new PR.

Once this PR is approved and merged will close
https://github.com/Azure/sonic-linux-kernel/pull/226

lguohan (approved):

PR #239, opened 2021-10-19
fix bullseye build for arm64 and armhf

lguohan:
suggest to use rebase to merge to keep individual commits


saiarcot895:
Is the original error message available? I'm curious why there's a build error for us on arm64, and the patch to fix the build error references upcoming changes to the networking headers.

lguohan:
they are in the individual commit message.

lguohan:
for arm64

Kernel support for 32-bit EL0 (COMPAT) [Y/n/?] y
  Enable kuser helpers page for 32-bit applications (KUSER_HELPERS) [Y/n/?] y
  Enable vDSO for 32-bit applications (COMPAT_VDSO) [Y/n/?] (NEW)

make -C olddefconfig will clear the this kconfig. thus, introduce
force_inclusion file and directly add the kconfig without further check.

saiarcot895 (approved):

PR #225, opened 2021-07-01
NPU disable unused PCIe BAR

msiddare:
NPU device enumeration was failing because of an issue in the cisco
ASIC. This patch will fix address the issue by disabling unused BAR.

Signed-off-by: Madhava Reddy Siddareddygari <msiddare@cisco.com>

PR #248, opened 2021-11-19
NPU disabled used PCIe BAR

msiddare:
For Cisco ASIC only BAR0 is valid. Not disabling other BAR's
was resulting in pci_enable_device function failure in P0
Pacific ASIC's. Further debugging and consultion with Hardware
team, issue seems to be related to only P0 version of ASIC and
workaround suggested is to disable unused PCI BAR.

This is same pull request reviewed ealier,
https://github.com/Azure/sonic-linux-kernel/pull/225/

Cisco experimented removing this patch, but identified that
there are boards in the field with Pacific ASIC.

This patch is needed for these ASIC's to work, hence re-submitting.

NPU is commonly used name for the packet forwarding ASIC's.

### DIFF
--- a/patch/cisco-npu-disable-other-bars.patch
+++ b/patch/cisco-npu-disable-other-bars.patch
@@ -1,24 +1,29 @@
-From d9ddce0a1c4699cf6c2c139da907a09cc1255178 Mon Sep 17 00:00:00 2001
+From 5b025a4732297147ec5c264225e1ffb50a09bfe5 Mon Sep 17 00:00:00 2001
 From: Madhava Reddy Siddareddygari <msiddare@cisco.com>
-Date: Mon, 15 Nov 2021 23:22:15 -0800
+Date: Wed, 1 Dec 2021 13:59:29 -0800
 Subject: [PATCH] NPU disable unused PCI BAR's
 
-For Cisco ASIC only BAR0 is valid. Not disabling other BAR's
-was resulting in pci_enable_device function failure in P0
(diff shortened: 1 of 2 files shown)


sachinv-msft:
@saiarcot895 fyi

paulmenzel (changes requested):
So this patch was dropped when moving from Linux 4.19 to 5.10? If so, please add that to the merge/pull request and commit message description.

Also, please add somewhere, what NPU means.

Some minor nit:

> This patch is need for these ASIC's to work, hence re-submitting.

is needed

patch/cisco-npu-disable-other-bars.patch
```diff
@@ -0,0 +1,71 @@
...
++{
++	int i;
++	struct resource *r;
++
++	for (i = 1; i <= PCI_ROM_RESOURCE; i++) {
++		r = &dev->resource[i];
++		if (!r->start && !r->end && !r->flags)
++			continue;
++
++		dev_info(&dev->dev, "Cisco Silicon One BAR %d %pR fixed up\n", i, r);
```
saiarcot895, line 46: Can you change `dev_info` to `pci_info`, for logging consistency with the rest of the file?

saiarcot895 (changes requested):

patch/cisco-npu-disable-other-bars.patch
msiddare: Updated patch with review comments

saiarcot895 (approved):

PR #250, opened 2021-12-07
drivers/pci: Reserve address space for devices behind bridges

msiddare:
Cisco data path devices are powered off by default, they will not be
available at BIOS stage and memory for these devices is not reserved.

This patch will reserve address space for data path devices that are
behind PCIe bridge, so that when devices are available PCIe subsystem
will be assign the address with in the specified range.

This patch is backported to 4.19.x kernel.


msiddare:
Closing this Pull request and will merge all Cisco back ports to a single pull request

PR #256, opened 2021-12-08
iio: (dac/ds4424) Allow NULL of_node

msiddare:

(cisco using ACPI for kernel module configuration). The kernel
(fw_node) in the device structure. It is unclear if they can be
using ACPI and not device tree. In this particular driver, there
reference to the device tree was made within the driver. This caused


paulmenzel:
So what is the difference to merge/pull request https://github.com/Azure/sonic-linux-kernel/pull/236?

msiddare:
Closing this Pull request and will merge all Cisco back ports to a single pull request

msiddare:
Closing this Pull request and will merge all Cisco back ports to a single pull request

msiddare:
Closing this Pull request and will merge all Cisco back ports to a single pull request

PR #260, opened 2022-02-09
Backport a patch to increase the GPIO count from 512 to 1024 on x86_64

saiarcot895:
Linux 5.15 bumped up the maximum GPIO count from the default of 512 to
1024 on x86_64 due to Tiger Lake-H SoC. Backport that patch from 5.15 to
replace the cisco-x86-gpio-config.patch.


saiarcot895:
FYI @madhava28 

This will get backported to 202012 branch.

patch/increase-maximum-gpio-number-for-x86-64.patch
```diff
@@ -0,0 +1,36 @@
+Backport from Linux 5.15
+
+commit d7109fe3a0991a0f7b4ac099b78c908e3b619787
+Author: Andy Shevchenko <andriy.shevchenko@linux.intel.com>
+Date:   Thu Aug 26 18:03:17 2021 +0300
+
+    x86/platform: Increase maximum GPIO number for X86_64
```
wenyiz2021, line 7: Hi Sai, May I know the purpose of changing file name?

wenyiz2021 (approved):

patch/increase-maximum-gpio-number-for-x86-64.patch
paulmenzel: It matches the git commit message summary.

patch/increase-maximum-gpio-number-for-x86-64.patch
```diff
@@ -0,0 +1,36 @@
+Backport from Linux 5.15
+
+commit d7109fe3a0991a0f7b4ac099b78c908e3b619787
```
paulmenzel, line 3: Please do not use the output of `git show` but create the patch with `git format-patch -1 d7109fe3a0991a0f7b4ac099b78c908e3b619787`, so it can be easily applied with `git am`.

PR #268, opened 2022-02-15
[202012] Apply kernel patches to fix emmc unreliability

Staphylo:
Some patches are required to fix some emmc unreliability seen on AMD platforms.
These were ported over from the EOS kernel and applied properly.

### DIFF
--- a/patch/driver-arista-mmcblk-not-working-on-AMD-platforms.patch
+++ b/patch/driver-arista-mmcblk-not-working-on-AMD-platforms.patch
@@ -1,0 +1,35 @@
+amd/mmc: mmcblk not working on some AMD platforms
+
+ADMA and ADMA-64 seem to be broken on AMD. This patch enables the
+following quirks (for AMD only):
+    SDHCI_QUIRK_BROKEN_ADMA
+    SDHCI_QUIRK2_BROKEN_64_BIT_DMA
+
+Signed-off-by: Radu Rendec <rrendec@arista.com>
+Signed-off-by: Samuel Angebault <staphylo@arista.com>
+---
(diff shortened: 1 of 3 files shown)


paulmenzel:
1.  Please document the corresponding master branch commit?
2.  Are these patches from upstream? If not, where are they from?

sujinmkang (approved):

Staphylo:
Hi Paul,

There are no corresponding master branch commits for these 2 patches yet, which is why there is no references to them.
The fixes for 4.9 (201811 / 201911) was different and more involved. It also had to be addressed in priority.
The fixes for 4.19 (202012 / 202106) and 5.10 (202111, master) are similar but I prefer patching them with dedicated PRs.

One of the patch adds an harmless cmdline parameter that allows configuring some extra sdhci quirks without overriding the ones autodetected.
The other one adds some quirks and quirks2 for older generation AMD CPUs.

These patches come from EosKernel which is the kernel used in EOS.
They have been thoroughly tested there and applied cleanly on sonic-linux-kernel.

Thanks,

paulmenzel:
Thank you for the explanation. I do not know, what the EOS kernel is.

Normally, I prefer the Linux policy, that all changes need to be in master first, before they get backported. Otherwise, it get’s too complicated to keep track, what should work in what version.

Staphylo:
EOS is Arista's operating system and EosKernel the kernel it uses.
It's a standard linux kernel with a few patches applied to it, similar to how sonic-linux-kernel works.
We aim to upstream/backport a maximum of patches to reduce the maintenance burden.

I do agree with you and that's my preferred path as well. I made an exception in this case due to priority.

PR #269, opened 2022-02-17
Increase log_buf_len to 1M on all architectures

Staphylo:
This increase allows for more logs to be stored in the kernel log buffer.
The default was 128k which this config change bumps to 1M.
It can prove useful in cases where /var/log becomes un-writtable.


prgeor:
@Staphylo in which scenario  "/var/log becomes un-writtable" ?

PR #271, opened 2022-03-01
[201911] Support vrf mgmt for aaa

inspurSDN:
Consider if the socket is bound to a device when binding.


patch/series
```diff
@@ -132,6 +132,7 @@ driver-ixgbe-external-phy.patch
 fix_ismt_alignment_issue.patch
 net-psample-module-unload.patch
 config-sensor-mp2975.patch
+Support-aaa-for-mgmt-vrf.patch
```
paulmenzel, line 135: Where is the patch file? Did you forget to add it with `git add`?

    stg import: [Errno 2] No such file or directory: '/__w/3/s/patch/Support-aaa-
                for-mgmt-vrf.patch'

paulmenzel (changes requested):
Please elaborate on the problem, and what devices you see this with.

Is that patch also part of the main branch? If not, please send it first to the main branch.

saiarcot895:
Why is this patch needed? Why is a security check being loosened?

PR #272, opened 2022-03-02
[Marvell-armhf] Updated kconfig exclusions

pavannaregundi:
Change-Id: I54e39017db276bbceabcf181b530537e5de2977b
Signed-off-by: Pavan Naregundi <pnaregundi@marvell.com>

kconfig exclusions necessary for resolving kernel boot hung issue for armhf.

Kernel hang in during early boot is caused due to CONFIG_ARCH_NR_GPIO is not set to 0 even after force inclusion file had this change. With further debugging found the root cause that CONFIG_ARCH_NR_GPIO is not getting set to 0 due to some dependent configs, which are now added as exclusions in this change.


saiarcot895:
In addition to what Paul said, one question I have is that the current value of `CONFIG_ARCH_NR_GPIO` appears to be 2048 for armhf (based on the most recent CI build on the master branch). That's more than the default of 512 that would get set in `include/asm-generic/gpio.h`, so that should be good enough for most systems. Is there something that I'm missing?

saiarcot895:
Also, if `CONFIG_ARCH_NR_GPIO` really needs to be user-configurable, then I'd recommend making a change similar to bb6ccd59687b3b113ac661cd3a4fe0caeaf7be4e instead of using kconfig-exclusions to exclude more kernel config settings.

pavannaregundi:
Thanks for review comments.
With further debugging, found root-cause of the kernel hang to be related to size of the kernel image built, and this config removal is just helping in reducing the kernel image size. With this analysis we find current PR is not relevant and we will open a new PR with a proper fix. Hence closing the PR. 

-Pavan

PR #278, opened 2022-06-02
[201911][patch] mlxsw: i2c: Prevent transaction execution for special chip st…

stepanblyschak:
…ates

Do not run transaction in cases chip is in reset or in-filed update states.

This patch fixes an issue encountered during SONiC warm-reboot: during
ISSU start the kernel reports that CPU stall is detected:

```
INFO: rcu_sched detected stalls on CPUs/tasks
```

This patch won't be upstreamed due to the upstream driver not supporting
ISSU.

Signed-off-by: Stepan Blyschak <stepanb@nvidia.com>

### DIFF
--- a/patch/0080-TMP-mlxsw-i2c-Prevent-transaction-execution-for-spec.patch
+++ b/patch/0080-TMP-mlxsw-i2c-Prevent-transaction-execution-for-spec.patch
@@ -1,0 +1,125 @@
+From 392127480c91a26dae2449288229d4957213dcc1 Mon Sep 17 00:00:00 2001
+From: Vadim Pasternak <vadimp@nvidia.com>
+Date: Tue, 31 May 2022 10:27:38 +0300
+Subject: [PATCH v4.9 ISSU WA 1/1] TMP: mlxsw: i2c: Prevent transaction
+ execution for special chip states
+
+Do not run transaction in cases chip is in reset or in-service update
+states.
+In such case firmware is not accessible and will reject transaction
+with the relevant status "RUNNING_RESET" or "FW_ISSU_ONGOING".
+In case transaction is failed do to one of these reasons, stop sending
+transactions. In such case driver is about to be removed since it
+cannot continue running after reset or in-service update. And
+re-probed again after reset or in-service update is completed.
+
+Signed-off-by: Vadim Pasternak <vadimp@nvidia.com>
+---
+ drivers/net/ethernet/mellanox/mlxsw/cmd.h |  4 ++++
+ drivers/net/ethernet/mellanox/mlxsw/i2c.c | 29 ++++++++++++++++++++---
+ 2 files changed, 30 insertions(+), 3 deletions(-)
+
+diff --git a/drivers/net/ethernet/mellanox/mlxsw/cmd.h b/drivers/net/ethernet/mellanox/mlxsw/cmd.h
+index 0772e4339b33..6329ebb09523 100644
+--- a/drivers/net/ethernet/mellanox/mlxsw/cmd.h
++++ b/drivers/net/ethernet/mellanox/mlxsw/cmd.h
+@@ -149,6 +149,8 @@ enum mlxsw_cmd_status {
+ 	MLXSW_CMD_STATUS_BAD_NVMEM	= 0x0B,
+ 	/* Device is currently running reset */
+ 	MLXSW_CMD_STATUS_RUNNING_RESET	= 0x26,
++	/* FW ISSU ongoing. */
++	MLXSW_CMD_STATUS_FW_ISSU	= 0x27,
+ 	/* Bad management packet (silently discarded). */
+ 	MLXSW_CMD_STATUS_BAD_PKT	= 0x30,
+ };
+@@ -180,6 +182,8 @@ static inline const char *mlxsw_cmd_status_str(u8 status)
+ 		return "BAD_NVMEM";
+ 	case MLXSW_CMD_STATUS_RUNNING_RESET:
+ 		return "RUNNING_RESET";
++	case MLXSW_CMD_STATUS_FW_ISSU:
++		return "FW_ISSU_ONGOING";
+ 	case MLXSW_CMD_STATUS_BAD_PKT:
+ 		return "BAD_PKT";
+ 	default:
+diff --git a/drivers/net/ethernet/mellanox/mlxsw/i2c.c b/drivers/net/ethernet/mellanox/mlxsw/i2c.c
+index e04d521d9376..1205c6111b0b 100644
+--- a/drivers/net/ethernet/mellanox/mlxsw/i2c.c
++++ b/drivers/net/ethernet/mellanox/mlxsw/i2c.c
+@@ -62,6 +62,7 @@
+  * @core: switch core pointer;
+  * @bus_info: bus info block;
+  * @block_size: maximum block size allowed to pass to under layer;
++ * @status: status to indicate chip reset or in-service update;
+  */
+ struct mlxsw_i2c {
+ 	struct {
+@@ -75,6 +76,7 @@ struct mlxsw_i2c {
+ 	struct mlxsw_core *core;
+ 	struct mlxsw_bus_info bus_info;
+ 	u16 block_size;
++	u8 status;
+ };
+
+ #define MLXSW_I2C_READ_MSG(_client, _addr_buf, _buf, _len) {	\
+@@ -221,6 +223,19 @@ static int mlxsw_i2c_write_cmd(struct i2c_client *client,
+ 	return 0;
+ }
+
++static bool
++mlxsw_i2c_cmd_status_verify(struct device *dev, struct mlxsw_i2c *mlxsw_i2c,
++			    u8 status)
++{
++	if (status == MLXSW_CMD_STATUS_FW_ISSU ||
++	    status == MLXSW_CMD_STATUS_RUNNING_RESET) {
++		mlxsw_i2c->status = status;
++		dev_info(dev, "FW status=%x(%s)): Access to device is not allowed in this state\n", status, mlxsw_cmd_status_str(status));
++		return true;
++	}
++	return false;
++}
++
+ /* Routine posts initialization command to ASIC through mail box. */
+ static int
+ mlxsw_i2c_write_init_cmd(struct i2c_client *client,
+@@ -404,6 +419,10 @@ mlxsw_i2c_cmd(struct device *dev, u16 opcode, u32 in_mod, size_t in_mbox_size,
+
+ 	WARN_ON(in_mbox_size % sizeof(u32) || out_mbox_size % sizeof(u32));
+
++	/* Do not run transaction if chip is in reset or in-service update state. */
++	if (mlxsw_i2c->status)
++		return 0;
++
+ 	if (in_mbox) {
+ 		reg_size = mlxsw_i2c_get_reg_size(in_mbox);
+ 		num = reg_size / mlxsw_i2c->block_size;
+@@ -478,6 +497,8 @@ mlxsw_i2c_cmd(struct device *dev, u16 opcode, u32 in_mod, size_t in_mbox_size,
+
+ cmd_fail:
+ 	mutex_unlock(&mlxsw_i2c->cmd.lock);
++	if (mlxsw_i2c_cmd_status_verify(&client->dev, mlxsw_i2c, *status))
++		err = 0;
+ 	return err;
+ }
+
+@@ -607,14 +628,16 @@ static int mlxsw_i2c_probe(struct i2c_client *client,
+ 	/* Wait until go bit is cleared. */
+ 	err = mlxsw_i2c_wait_go_bit(client, mlxsw_i2c, &status);
+ 	if (err) {
+-		dev_err(&client->dev, "HW semaphore is not released");
++		if (!mlxsw_i2c_cmd_status_verify(&client->dev, mlxsw_i2c, status))
++			dev_err(&client->dev, "HW semaphore is not released");
+ 		goto errout;
+ 	}
+
+ 	/* Validate transaction completion status. */
+ 	if (status) {
+-		dev_err(&client->dev, "Bad transaction completion status %x\n",
+-			status);
++		if (!mlxsw_i2c_cmd_status_verify(&client->dev, mlxsw_i2c, status))
++			dev_err(&client->dev, "Bad transaction completion status %x\n",
++				status);
+ 		err = -EIO;
+ 		goto errout;
+ 	}
+--
+2.20.1
--- a/patch/0080-mlxsw-i2c-Prevent-transaction-execution-for-spec.patch
+++ b/patch/0080-mlxsw-i2c-Prevent-transaction-execution-for-spec.patch
@@ -1,7 +1,7 @@
 From 392127480c91a26dae2449288229d4957213dcc1 Mon Sep 17 00:00:00 2001
 From: Vadim Pasternak <vadimp@nvidia.com>
 Date: Tue, 31 May 2022 10:27:38 +0300
-Subject: [PATCH v4.9 ISSU WA 1/1] TMP: mlxsw: i2c: Prevent transaction
+Subject: [PATCH v4.9 ISSU WA 1/1] mlxsw: i2c: Prevent transaction
  execution for special chip states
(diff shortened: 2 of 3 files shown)


lguohan (approved):

patch/0080-TMP-mlxsw-i2c-Prevent-transaction-execution-for-spec.patch
```diff
@@ -0,0 +1,82 @@
+From b041f1a4ed2da36a831548d8539e8869cea33b40 Mon Sep 17 00:00:00 2001
+From: Vadim Pasternak <vadimp@nvidia.com>
+Date: Tue, 31 May 2022 10:27:38 +0300
+Subject: [PATCH v4.9 ISSU WA 1/1] TMP: mlxsw: i2c: Prevent transaction
+ execution for special chip states
+
+Do not run transaction in cases chip is in reset or in-filed update states.
+
```
paulmenzel, line 8: Is that commit upstream? Plesae explain the TMP.

paulmenzel (changes requested):
Please document the problem, and why it’s needed only for the 201911 branch?

patch/0080-TMP-mlxsw-i2c-Prevent-transaction-execution-for-spec.patch
```diff
@@ -0,0 +1,82 @@
...
+ 	if (in_mbox) {
+ 		reg_size = mlxsw_i2c_get_reg_size(in_mbox);
+ 		num = reg_size / mlxsw_i2c->block_size;
+@@ -478,6 +484,11 @@ mlxsw_i2c_cmd(struct device *dev, u16 opcode, u32 in_mod, size_t in_mbox_size,
+
+ cmd_fail:
+ 	mutex_unlock(&mlxsw_i2c->cmd.lock);
++	if (*status == MLXSW_CMD_STATUS_FW_ISSU ||
++	    *status == MLXSW_CMD_STATUS_RUNNING_RESET) {
++		mlxsw_i2c->status = *status;
```
saiarcot895, line 74: Is it assumed that after it gets into either `MLXSW_CMD_STATUS_FW_ISSU` or `MLXSW_CMD_STATUS_RUNNING_RESET` state, it won't get out of this state or there will be a kexec to a new kernel/image? Otherwise, unless I'm missing something, future i2c commands won't get executed.

patch/0080-TMP-mlxsw-i2c-Prevent-transaction-execution-for-spec.patch
```diff
@@ -0,0 +1,82 @@
...
++		return 0;
++
+ 	if (in_mbox) {
+ 		reg_size = mlxsw_i2c_get_reg_size(in_mbox);
+ 		num = reg_size / mlxsw_i2c->block_size;
+@@ -478,6 +484,11 @@ mlxsw_i2c_cmd(struct device *dev, u16 opcode, u32 in_mod, size_t in_mbox_size,
+
+ cmd_fail:
+ 	mutex_unlock(&mlxsw_i2c->cmd.lock);
++	if (*status == MLXSW_CMD_STATUS_FW_ISSU ||
```
vaibhavhd, line 72: Also, in what scenario will the status be `MLXSW_CMD_STATUS_RUNNING_RESET`?
I understand why ISSU (warmboot) case is covered here, but don't understand why RESET is also covered.

patch/0080-TMP-mlxsw-i2c-Prevent-transaction-execution-for-spec.patch
liat-grozovik: it is not and it cannot be upstream. this is a way for bug fix and we wont be able to upstream it.

patch/0080-TMP-mlxsw-i2c-Prevent-transaction-execution-for-spec.patch
vadimp-nvidia: In case of reset I2C chip driver should be removed and re-probed. Since state check for ISSU is added it's logical alos to handle reset state as well.

patch/0080-TMP-mlxsw-i2c-Prevent-transaction-execution-for-spec.patch
```diff
@@ -0,0 +1,82 @@
...
++	MLXSW_CMD_STATUS_FW_ISSU	= 0x27,
+ 	/* Bad management packet (silently discarded). */
+ 	MLXSW_CMD_STATUS_BAD_PKT	= 0x30,
+ };
+@@ -180,6 +182,8 @@ static inline const char *mlxsw_cmd_status_str(u8 status)
+ 		return "BAD_NVMEM";
+ 	case MLXSW_CMD_STATUS_RUNNING_RESET:
+ 		return "RUNNING_RESET";
++	case MLXSW_CMD_STATUS_FW_ISSU:
++		return "FW_ISSUE_ONGOING";
```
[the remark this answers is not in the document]
vadimp-nvidia: Should be fixed.

patch/0080-TMP-mlxsw-i2c-Prevent-transaction-execution-for-spec.patch
vadimp-nvidia: The intention is to not run any I2C transaction ASIC in cases chip reset or in-service update states is detected.
Firmware is not accessible and will reject transaction with the relevant status "RUNNING_RESET" or "FW_ISSU_ONGOING".
In case transaction is failed do to one of these reasons, stop sending transactions.
In such state driver is about to be removed since it cannot continue running after reset or in-service update.
And re-probed again after reset or in-service update is completed.

For ISSU system will be rebooted by kexec.
For chip reset ASIC I2C driver is handled in user space by system udev event (removing/probing).

patch/0080-TMP-mlxsw-i2c-Prevent-transaction-execution-for-spec.patch
paulmenzel: I do not understand. Please elaborate.

vadimp-nvidia: Upstream mlxsw driver doesn't support ISSU, so we can't submit ISSU related filtering to upstream.

patch/0080-TMP-mlxsw-i2c-Prevent-transaction-execution-for-spec.patch
vaibhavhd: I also want to make sure that this is also covering the non-reboot scenarios - config-reload, cold restart of syncd/swss etc.
Please confirm that after these non-reboot actions, the i2c commands still get executed.

patch/0080-TMP-mlxsw-i2c-Prevent-transaction-execution-for-spec.patch
paulmenzel: Only 4.19(?) or all versions? The policy of SONiC is that long term all patches have to be upstream to lessen the maintenance burden especially when updating to newer Linux kernel releases.

Can you please donate that in the patch description?

liat-grozovik:
Please dont merge yet.
We should run also fast and cold reboot tests to ensure these reboots are also clean with the fix.
Please wait for Nvidia/me confirmation on it.

vaibhavhd:
> Please dont merge yet. We should run also fast and cold reboot tests to ensure these reboots are also clean with the fix. Please wait for Nvidia/me confirmation on it.

Perhaps also add config-reload, load_minigraph or any other scenario (syncd cold reboot) which might trigger this.

liat-grozovik:
@roysr-nv , @stepanblyschak  once all tests are done and we are good to go with the fix please move it to ready for review.

paulmenzel (changes requested):
Please squash both commits, and add more information for the ignorant, that ISSU is not supported upstream (all versions), and when upstreaming is going to happen.

stepanblyschak:
@paulmenzel Updated per your request

patch/0080-TMP-mlxsw-i2c-Prevent-transaction-execution-for-spec.patch
```diff
@@ -0,0 +1,126 @@
...
+ }
+
++static bool
++mlxsw_i2c_cmd_status_verify(struct device *dev, struct mlxsw_i2c *mlxsw_i2c,
++			    u8 status)
++{
++	if (status == MLXSW_CMD_STATUS_FW_ISSU ||
++	    status == MLXSW_CMD_STATUS_RUNNING_RESET) {
++		mlxsw_i2c->status = status;
++		dev_info(dev, "FW status=%x(%s))\n", status, mlxsw_cmd_status_str(status));
```
paulmenzel, line 75: This looks more like a debug message instead of an informational message. Could you please elaborate/improve the wording?

patch/0080-TMP-mlxsw-i2c-Prevent-transaction-execution-for-spec.patch
```diff
@@ -0,0 +1,126 @@
+From 392127480c91a26dae2449288229d4957213dcc1 Mon Sep 17 00:00:00 2001
+From: Vadim Pasternak <vadimp@nvidia.com>
+Date: Tue, 31 May 2022 10:27:38 +0300
+Subject: [PATCH v4.9 ISSU WA 1/1] TMP: mlxsw: i2c: Prevent transaction
```
paulmenzel, line 4: Why the TMP tag?

paulmenzel (commented):
Thank you for updating the merge/pull request description. Please update also the commit message in the patch file.

patch/0080-TMP-mlxsw-i2c-Prevent-transaction-execution-for-spec.patch
vadimp-nvidia: Why? It will print something like
mlxsw_minimal 2-0048: FW status=27(FW_ISSU_ONGOING)

paulmenzel: What is a user (non-developer) supposed to do with that message?

vadimp-nvidia: Do you prefer something like:
mlxsw_minimal 2-0048: FW status=27(FW_ISSU_ONGOING): Access to device is not allowed in this state

paulmenzel: Yes, I would consider that an improvement.

stepanblyschak:
Updated with latest patch from @vadimp-nvidia

liat-grozovik:
@lguohan is there anything else open related this PR or we can approve and signoff?
once this is done we need to have it on all other branches. PRs are already avaialble.

patch/0080-TMP-mlxsw-i2c-Prevent-transaction-execution-for-spec.patch
saiarcot895: Is the TMP tag needed (as in, will this bug fix be removed in the future), or is it expected to be permanent?

vadimp-nvidia: It is expected to be permanent.
We want to mark it as downstream patch.
There are a lot of not up-streamed patches in Sonic kernel, But there is no convention in Sonic  to tag such patches.
"TMP" in patch name could be replaced with "DS" or "Downstream" or with something like that.

saiarcot895: I would rather have "Sonic-specific" or "Sonic-only" in the patch rather than TMP.

vadimp-nvidia: "Sonic-specific" sounds better to me. Can we close with this name?
I suggest to use this tag for all new-coming patches (not only from Nvidia), which are not going to be up-streamed.

patch/0080-TMP-mlxsw-i2c-Prevent-transaction-execution-for-spec.patch
paulmenzel: In the repository, I did `git grep -i issu`, and nothing showed up regarding *ISSU*. Where can I find more information?

patch/0080-TMP-mlxsw-i2c-Prevent-transaction-execution-for-spec.patch
paulmenzel: I believe such a tag is unnecessary. The README suggests, that patches are not wanted, that are not going to be upstreamed. The only exception is:

> Platform specific kernel modules which are impossible or very difficult to be built out of kernel tree.

Why were the patches accepted to be in SONiC in the first place. I really thought these were only backports, and it’s not made easy for reviewers to find information about the feature (ISSU), and why it’s not going to be upstreamed.

paulmenzel (changes requested):

patch/0080-TMP-mlxsw-i2c-Prevent-transaction-execution-for-spec.patch
vadimp-nvidia: There is no ISSU feature in up-stream driver, so this patch can't be accepted.

patch/0080-TMP-mlxsw-i2c-Prevent-transaction-execution-for-spec.patch
saiarcot895: Within SONiC, ISSU may be called warm reboot or warmboot. Some more details are [here](https://github.com/sonic-net/SONiC/blob/master/doc/warm-reboot/SONiC_Warmboot.md).

patch/0080-TMP-mlxsw-i2c-Prevent-transaction-execution-for-spec.patch
saiarcot895: I agree that there are a number of patches here that should be upstreamed. However, there can still be patches that are not useful to be upstreamed, such as patches that would break the general usage of the Linux kernel, or some hack that really shouldn't be upstreamed.

In this case, since this is functionality that is not supported by the upstream driver, it wouldn't make sense to commit this bug fix for functionality that's not even present in the upstream Linux kernel. If this changes (i.e. ISSU/warmboot/equivalent functionality) becomes available in the upstream kernel, then this should be upstreamed as well.

patch/0080-TMP-mlxsw-i2c-Prevent-transaction-execution-for-spec.patch
paulmenzel: Thank you: ISSU = In-Service-Software-Upgrade

patch/0080-TMP-mlxsw-i2c-Prevent-transaction-execution-for-spec.patch
paulmenzel: Thank you, but I still do not understand. Does SONiC add a special module/patch to support ISSU in it’s Linux kernel build? Where are those patches? I could apply this patch here (for current SONiC branch) to the upstream Linux kernel source just fine.

    $ git am ~/src/sonic-linux-kernel/patch/0064-TMP-mlxsw-i2c-Prevent-transaction-execution-for.patch # nvidia-stepan/202106-prevent-transaction
    Wende an: TMP: mlxsw: i2c: Prevent transaction execution for special chip states

`MLXSW_CMD_STATUS_FW_ISSU` sounds to me like it’s support in the firmware and independent from the Linux kernel, so should be submitted upstream.

vadimp-nvidia: Since upstream doesn’t support ISSU, it means nobody can set FW to ISSU and FW will never get MLXSW_CMD_STATUS_FW_ISSU. That’s we it can’t be up-streamed.

stepanblyschak:
@paulmenzel @saiarcot895 @vadimp-nvidia  I removed the TMP tag from the commit message

PR #282, opened 2022-06-23
[202012][patch] mlxsw: i2c: Prevent transaction execution for special

stepanblyschak:
… chip

states

Do not run transaction in cases chip is in reset or in-filed update states.

This patch fixes an issue encountered during SONiC warm-reboot: during
ISSU start the kernel reports that CPU stall is detected:

```
INFO: rcu_sched detected stalls on CPUs/tasks
```

This patch won't be upstreamed due to the upstream driver not supporting
ISSU.

Signed-off-by: Stepan Blyschak <stepanb@nvidia.com>


stepanblyschak:
@saiarcot895 Could you please also review this PR for 202012?

saiarcot895 (approved):

PR #286, opened 2022-07-20
Remove kernel config SERIAL_8250_RUNTIME_UARTS

madhava28:
Issue:
UART parameter of Cisco linux kernel configuration will cause Accton platform
not to dump any message on booting stage.

Fix:
Remove SERIAL_8250_RUNTIME_UARTS config. This will be passed as bootargs to the
kernel during bootup.


lguohan (approved):

paulmenzel:
Please elaborate the problem. What Accton platform fails to boot, and what is the reason.

What is now happening with the Cisco devices? Will they still work? Why was the value set to 32 in the first place?

paulmenzel (changes requested):
Updating GitHub meta data.

lguohan:
here is the issue. https://github.com/Azure/sonic-buildimage/issues/11238

the plan is that cisco will revert this one, and then use bootargs to set it up only for cisco platforms.

paulmenzel (changes requested):
Thank you. Please amend the commit message with a summary of the issue in the commit message, add document Linux CLI parameter name.

saiarcot895 (approved):

PR #289, opened 2022-08-09
[201911][Mellanox]Adding mlsxw debug patch

dgsudharsan:
Adding a debug patch to retry i2c transactions on failure. 
Currently I2C transaction used to prepare control register for sending transaction to ASIC and setting "go" bit in this register are not re-tried. Add retries in case these transactions are replied with NACK.The purpose of this patch is for debugging I2C transaction failures found on some MSFT SN2700 systems.In case one of two kinds of transactions is failed, transaction is to
be re-tried until success or until both 5000 milliseconds time-out
window and retries counter 5 are expired.


paulmenzel:
Is that the same as merge/pull request https://github.com/sonic-net/sonic-linux-kernel/pull/290, and therefore can be closed?

PR #296, opened 2022-10-18
Update Makefile to provision the ability of building with non-upstream patches

vivekrnv:
Dependent PR: https://github.com/sonic-net/sonic-buildimage/pull/12428

Can be merged in any order

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -110,7 +110,7 @@ $(addprefix $(DEST)/, $(MAIN_TARGET)): $(DEST)/% :
 	# Precedence is given for external URL
 	if [ -z ${EXTERNAL_KERNEL_PATCHES} ] && [ x${INCLUDE_MLNX_PATCHES} == xy ]; then
 		if [ -f "$(MLNX_PATCH_LOC)" ]; then
-			tar $(MLNX_PATCH_LOC) -C $(NON_UP_LOC)
+			tar -zxf $(MLNX_PATCH_LOC) -C $(NON_UP_LOC)
 		fi
 	fi
 
--- a/Makefile
+++ b/Makefile
@@ -109,7 +109,7 @@ $(addprefix $(DEST)/, $(MAIN_TARGET)): $(DEST)/% :
 
 	# Precedence is given for external URL
 	if [ -z ${EXTERNAL_KERNEL_PATCHES} ] && [ x${INCLUDE_MLNX_PATCHES} == xy ]; then
-		if [ -f "../../../$(MLNX_PATCH_LOC)" ]; then
+		if [ -f "$(MLNX_PATCH_LOC)" ]; then
 			tar $(MLNX_PATCH_LOC) -C $(NON_UP_LOC)
 		fi
 	fi
--- a/Makefile
+++ b/Makefile
@@ -56,7 +56,7 @@ SOURCE_FILE_BASE_URL="https://sonicstorage.blob.core.windows.net/debian-security
 DSC_FILE_URL = "$(SOURCE_FILE_BASE_URL)/$(DSC_FILE)"
 DEBIAN_FILE_URL = "$(SOURCE_FILE_BASE_URL)/$(DEBIAN_FILE)"
 ORIG_FILE_URL = "$(SOURCE_FILE_BASE_URL)/$(ORIG_FILE)"
-NON_UP_LOC = non_upstream_patches
+NON_UP_DIR = /tmp/non_upstream_patches
 
 $(addprefix $(DEST)/, $(MAIN_TARGET)): $(DEST)/% :
 	# Obtaining the Debian kernel source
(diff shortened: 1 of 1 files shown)


paulmenzel (changes requested):
Thank you for the patches. Please amend the commit messages to follow [*How to Write a Git Commit Message*](https://gitlab.freedesktop.org/drm/amd/-/issues/1915), and to clean it up.

Please also write a merge/pull request description describing the problem, and your solution. For example, it sounds like a directory `non_upstream_patches` is used by default. This has to be documented.

Lastly, what do you mean by *upstream*? https://www.kernel.org? Please clarify.

liat-grozovik:
@saiarcot895 please help to review

Makefile
```diff
@@ -99,6 +100,26 @@ $(addprefix $(DEST)/, $(MAIN_TARGET)): $(DEST)/% :
 	stg repair
 	stg import -s ../patch/series
 
+	mkdir $(NON_UP_LOC)
+
+	if [ ! -z ${EXTERNAL_KERNEL_PATCHES} ];  then
+		wget ${EXTERNAL_KERNEL_PATCHES} -O patches.tar.gz
+		tar -zxf patches.tar.gz -C $(NON_UP_LOC)
```
saiarcot895, line 107: Please use `tar -xf` instead of `tar -zxf` here and the one below, in accordance with [this](https://github.com/sonic-net/sonic-buildimage/pull/12428/files#r998395205) comment.

vivekrnv:
> Lastly, what do you mean by _upstream_? https://www.kernel.org? Please clarify.

It means patches that are not-yet or waiting to be upstreamed to mainline linux kernel. For a patch to be included in this repo, it has to be upstreamed in the mainline. This solution helps building the kernel with non-upstream patches (if any required)

saiarcot895 (approved):

Makefile
```diff
@@ -99,6 +100,26 @@ $(addprefix $(DEST)/, $(MAIN_TARGET)): $(DEST)/% :
 	stg repair
 	stg import -s ../patch/series
 
+	mkdir $(NON_UP_LOC)
```
saiarcot895, line 103: Can you add a line to remove this directory before creating it, in case it exists already?

Makefile
```diff
@@ -56,6 +56,7 @@ SOURCE_FILE_BASE_URL="https://sonicstorage.blob.core.windows.net/debian-security
 DSC_FILE_URL = "$(SOURCE_FILE_BASE_URL)/$(DSC_FILE)"
 DEBIAN_FILE_URL = "$(SOURCE_FILE_BASE_URL)/$(DEBIAN_FILE)"
 ORIG_FILE_URL = "$(SOURCE_FILE_BASE_URL)/$(ORIG_FILE)"
+NON_UP_LOC = non_upstream_patches
```
saiarcot895, line 59: Can you rename this to `NON_UP_DIR`, to make it clear it's a directory? Also, this directory should be outside of the kernel source code, so that there's no chance of this directory conflicting with anything coming from the kernel (or Debian) itself.

Makefile
vivekrnv: This sits under BUIlD_DIR and we delete that in the beginning, so this isn’t a problem

Makefile
```diff
@@ -99,6 +100,26 @@ $(addprefix $(DEST)/, $(MAIN_TARGET)): $(DEST)/% :
...
 
+	mkdir $(NON_UP_LOC)
+
+	if [ ! -z ${EXTERNAL_KERNEL_PATCHES} ];  then
+		wget ${EXTERNAL_KERNEL_PATCHES} -O patches.tar
+		tar -xf patches.tar -C $(NON_UP_LOC)
+	fi
+
+	# Precedence is given for external URL
+	if [ -z ${EXTERNAL_KERNEL_PATCHES} ] && [ x${INCLUDE_MLNX_PATCHES} == xy ]; then
```
saiarcot895, line 111: Thinking about the current kernel build, the kernel build is not platform-specific, and I'd rather not have platform-specific references in the build scripts. There's also a bit of a naming issue, where `EXTERNAL_KERNEL_PATCHES` could be set, but `INCLUDE_MLNX_PATCHES` could be `n`, and the patches would get applied, which seems counter-intuitive.

For the variable names, please don't have any platform-specific references within this file. Also, to make it a bit more clear, it should be something like this:
* If `INCLUDE_EXTERNAL_PATCHES` is not `y`, then nothing new happens; otherwise,
* If `EXTERNAL_KERNEL_PATCH_URL` is set, then the tarball is downloaded and extracted
* If `EXTERNAL_KERNEL_PATCH_TAR` is set, then the tarball is extracted.

saiarcot895 (changes requested):
New comments

Makefile
vivekrnv: How about the minor change:
* If `EXTERNAL_KERNEL_PATCH_URL` is set by the user, then the tarball is downloaded and extracted
* If `INCLUDE_EXTERNAL_PATCHES=y` is set by the user the code uses `EXTERNAL_KERNEL_PATCH_TAR` which is supposed to be set in the platform/ dir and then the tarball is extracted using that absolute path. `EXTERNAL_KERNEL_PATCH_TAR` shouldn't be passed from user and must be hardcoded and is specific to the platform.

saiarcot895: That sounds good.

Makefile

Makefile
vivekrnv: Updated

saiarcot895 (approved):

PR #298, opened 2022-11-13
Add Secure Boot Kernel configuration

davidpil2002:
In order to support the Secure Boot feature it required some modifications when building the Linux Kernel. 
This PR contained the kernel configuration aggregations to support it.

sonic-buildimage PR link: https://github.com/sonic-net/sonic-buildimage/pull/12692
HLD: https://github.com/sonic-net/SONiC/pull/1028


Makefile
```diff
@@ -126,6 +126,19 @@ $(addprefix $(DEST)/, $(MAIN_TARGET)): $(DEST)/% :
...
 	fi
 
+ifeq ($(SECURE_UPGRADE_MOD),$(filter $(SECURE_UPGRADE_MOD),dev prod))
+	if [ -f $(SECURE_UPGRADE_DEV_SIGNING_CERT) ]; then
+		echo "Add secure boot support in kernel config file"
+		cp ../patch/secure_boot_kernel_config.sh .
+		cp $(SECURE_UPGRADE_DEV_SIGNING_CERT) debian/certs
+		bash secure_boot_kernel_config.sh $(SECURE_UPGRADE_DEV_SIGNING_CERT)
+	else
+		echo "no certificate file exist, SECURE_UPGRADE_DEV_SIGNING_CERT=$(SECURE_UPGRADE_DEV_SIGNING_CERT)"
```
paulmenzel, line 136: exist*s*

patch/secure_boot_kernel_config.sh
```diff
@@ -0,0 +1,54 @@
+#!/bin/bash
```
paulmenzel, line 1: Can just `/bin/sh` be used?

patch/secure_boot_kernel_config.sh
```diff
@@ -0,0 +1,54 @@
...
+
+# Note: this script was created because there is a problem when changing the kernel config
+# values that requires in the Secure Boot feature when using patch/kconfig-inclusions (sonic flow to modify kernel flags).
```
paulmenzel, line 4: s/that requires in/required by/

patch/secure_boot_kernel_config.sh
```diff
@@ -0,0 +1,54 @@
...
+# So, when this problem will be resolved, this script should be removed and used the kconfig-inclusions.
```
paulmenzel, line 5: Where is that problem documented? Please open an issue with the details.

patch/secure_boot_kernel_config.sh
```diff
@@ -0,0 +1,54 @@
...
+$0 <PEM_CERT>
+Script is modifying kernel config file to support system trusted key with custom certificate.
+Note: The signature algorithem used will be SHA512.
+
+Parameters description:
+PEM_CERT                             public key (pem format). Key to be store in kernel.
+
+Run example:
+bash secure_boot_kernel_config.sh cert.pem
+EOF
```
paulmenzel, line 19: Can’t normal `echo` be used?

patch/secure_boot_kernel_config.sh
```diff
@@ -0,0 +1,54 @@
...
+# Note: this script was created because there is a problem when changing the kernel config
+# values that requires in the Secure Boot feature when using patch/kconfig-inclusions (sonic flow to modify kernel flags).
+# So, when this problem will be resolved, this script should be removed and used the kconfig-inclusions.
+
+usage() {
+    cat <<EOF
+$0: # Display Help
+$0 <PEM_CERT>
+Script is modifying kernel config file to support system trusted key with custom certificate.
+Note: The signature algorithem used will be SHA512.
```
paulmenzel, line 12: algorithm

patch/secure_boot_kernel_config.sh
```diff
@@ -0,0 +1,54 @@
...
+sed -i 's/^CONFIG_SECURITY_LOCKDOWN_LSM_EARLY=.*/# CONFIG_SECURITY_LOCKDOWN_LSM_EARLY is not set/g' $linux_cfg_file
+sed -i 's/^CONFIG_LOCK_DOWN_KERNEL_FORCE_NONE=.*/# CONFIG_LOCK_DOWN_KERNEL_FORCE_NONE is not set/g' $linux_cfg_file
+sed -i 's/^CONFIG_LOCK_DOWN_IN_EFI_SECURE_BOOT=.*/# CONFIG_LOCK_DOWN_IN_EFI_SECURE_BOOT is not set/g' $linux_cfg_file
+
+# warm boot secure
+sed -i 's/# CONFIG_KEXEC_SIG_FORCE is not set/CONFIG_KEXEC_SIG_FORCE=y/g' $linux_cfg_file
+
+echo "$0: Secure Boot support in Kernel config file DONE."
+
+
```
paulmenzel, line 54: Please remove trailing blank lines.

paulmenzel (changes requested):
Please amend the commit message, detailing the Secure Boot problem, how it was tested tested (also QEMU?), and why it can’t be enabled unconditionally.

Please add a space after ].

patch/secure_boot_kernel_config.sh
davidpil2002: format X.509, RSA algo, I will add it to the code description

Makefile
davidpil2002: fixed

patch/secure_boot_kernel_config.sh
davidpil2002: fix typo

patch/secure_boot_kernel_config.sh
davidpil2002: its also common to add some "Usage example:" as well

patch/secure_boot_kernel_config.sh
davidpil2002: done

davidpil2002 (commented):
fix few typos

patch/secure_boot_kernel_config.sh
davidpil2002: maybe sh can be use, but I just tested it with bash, in general in sonic-slave "sh" is a soft link to "dash".
`lrwxrwxrwx 1 root root 4 Oct 24 00:00 /bin/sh -> dash
`
And most of the scripts are using bash, so I think is fine.

patch/secure_boot_kernel_config.sh
davidpil2002: https://github.com/sonic-net/sonic-linux-kernel/issues/299

saiarcot895:
So Debian's/Ubuntu's approach to supporting Secure Boot is to sign the kernel after it's built with a shim binary. Could that work here, or is this approach to instead support any arbitrary public key that might be present in the UEFI?

patch/secure_boot_kernel_config.sh
```diff
@@ -0,0 +1,51 @@
...
+}
+
+local_cert_pem="debian/certs/$(basename $CERT_PEM)"
+linux_cfg_file="debian/build/build_amd64_none_amd64/.config"
+sed -i "s|^CONFIG_SYSTEM_TRUSTED_KEYS=.*|CONFIG_SYSTEM_TRUSTED_KEYS=\"$local_cert_pem\"|g" $linux_cfg_file
+sed -i 's/^CONFIG_MODULE_SIG_HASH=.*/CONFIG_MODULE_SIG_HASH="sha512"/g' $linux_cfg_file
+sed -i 's/^CONFIG_MODULE_SIG_SHA256=.*/# CONFIG_MODULE_SIG_SHA256 is not set/g' $linux_cfg_file
+sed -i 's/# CONFIG_MODULE_SIG_SHA512 is not set/CONFIG_MODULE_SIG_SHA512=y/g' $linux_cfg_file
+
+#lockdown feature disable
```
saiarcot895, line 42: Why does lockdown need to be disabled?

saiarcot895:
Regarding the `manage-config` script issue, I've raised #300 to fix it.

davidpil2002:
> So Debian's/Ubuntu's approach to supporting Secure Boot is to sign the kernel after it's built with a shim binary. Could that work here, or is this approach to instead support any arbitrary public key that might be present in the UEFI?

yes,
We are doing the full Secure Boot flow like the Debian approach suggested, using shim, grub, and kernel signed, also the feature is signing kernel modules.
more details can be found in the HLD(link in the description)
And yes, by using shim you can store any arbitrary pub key in the UEFI DB

davidpil2002:
> Regarding the `manage-config` script issue, I've raised #300 to fix it.

Thanks!

patch/secure_boot_kernel_config.sh
davidpil2002: We are aware of this point, from the Linux Debian definitions it can be enabled,  but when setting the Secure Boot enabled in SONiC with LockDown it created some conflict, we are working on it.

davidpil2002:
> > Regarding the `manage-config` script issue, I've raised #300 to fix it.
> 
> Thanks!

liat-grozovik:
@saiarcot895  could you please add a label 'request for 202211'?

liat-grozovik:
@saiarcot895 can you please help to merge and label?

lguohan (changes requested):
the approach here is complicated, secure boot kernel configuration should be incorprated into manage-config script, for you can add option in manage-config to enable and disable secure boot and then have a secureboot_config to have those options to be included like other existing config.

liat-grozovik:
@davidpil2002  please update your branch to the latest
@saiarcot895 are you ok with the change and can approve it?

saiarcot895:
> @saiarcot895 are you ok with the change and can approve it?

No. I concur with Guohan's comment.

1. The changes that the new script is making results in modifying files that are tracked in the repo. This means after a build (or even attempted build, in case of a build failure), the git repo's state will be dirty (because there are changes to tracked files).
2. There's nothing undoing these changes after the build is done, meaning if someone switches from a secure boot to non-secure boot environment variable settings, then they'll need to reset the git repo as well.
3. Considering all that is being done is setting some configs (one of which takes in user input), having it be done in `manage-config` makes more sense IMO. `kconfig-secure-boot-inclusions` and `kconfig-secure-boot-exclusions` can be created as well to store the config changes that don't depend on user input.

saiarcot895:
> if there is a fail, the makefile always removes the Linux repo and does wget of the kernel repo fresh. So no see a problem here

I was referring to the changes done in these lines to `patch/kconfig-inclusions` and `patch/kconfig-exclusions`.

https://github.com/sonic-net/sonic-linux-kernel/blob/77e18425b9a1ad12cdcb8fc34c09640bea4bb92e/patch/secure_boot_kernel_config.sh#L60-L69

These files are part of the sonic-linux-kernel repo, rather than the Linux kernel repo.

> The SB depends on the user build configuration because the user needs to decide if SB is enabled or disabled at start of build time.
> So I am not sure 100% that creating kconfig-secure-boot-inclusions and kconfig-secure-boot-exclusions can solve the problem.
> The reason is because the kernel configuration that the feature required is not static, it depends on the user configuration selected at the start of the build time.
> Maybe until today, the sonic-Linux-kernel supported just static configuration and now we need a new solution that removes the Linux target if some user build flag related was changed?

I think there's two different things being touched on here. The purpose of creating `kconfig-secure-boot-inclusions` and `kconfig-secure-boot-exclusions` is to have just the static secure boot configs (i.e. everything except `CONFIG_SYSTEM_TRUSTED_KEYS`) in a regular plain-text file, which will then be read in by `manage-config` (similar to how it's reading in `kconfig-inclusions` and `kconfig-exclusions`) and apply the secure boot configs if some argument is passed in. `manage-config` would also separately set `CONFIG_SYSTEM_TRUSTED_KEYS` based on the argument value passed in.

For the kernel build from sonic-buildimage and controlling the configuration, whether the kernel will get rebuilt after it's been already built once depends on the `DEP_FLAGS` for the kernel build:

https://github.com/sonic-net/sonic-buildimage/blob/5e4a866e3311f859b9fe479aa379b38521de9725/rules/linux-kernel.dep#L6-L7

This variable would need to be updated to include the config variable that would be specified in sonic-buildimage to control whether secure boot is enabled and what certificate is used. That way, in a build with caching enabled, if someone switches between having secure boot enabled or disabled, or even different certificate paths, then the kernel will get rebuilt (I don't recall if this is fully checked or not in a build with caching disabled).

patch/kconfig-secure-boot-inclusions
```diff
@@ -0,0 +1,20 @@
+[common]
+
+[amd64]
+# CONFIG_SYSTEM_TRUSTED_KEYS="$LOCAL_CERT_PEM"
```
DavidZagury, line 4: Remove comment

patch/kconfig-secure-boot-inclusions
```diff
@@ -0,0 +1,20 @@
...
+CONFIG_SYSTEM_TRUSTED_KEYS="/sonic/files/nv_onyx_key_certificate.pem"
```
DavidZagury, line 5: Fix to use generic file

DavidZagury (changes requested):

manage-config
```diff
@@ -58,88 +67,113 @@ function get_section_opts(){
...
+#  Secure Boot support
+if [ $ret_process_inc_ex -e 0 ]; then
+    if [ ${SECURE_UPGRADE_MODE} == "dev" -o ${SECURE_UPGRADE_MODE} == "prod" ]; then
+        echo "set kconfig-secure-boot-exclusions & kconfig-secure-boot-inclusions"
+        exclusion_file="../patch/kconfig-secure-boot-exclusions"
+        inclusion_file="../patch/kconfig-secure-boot-inclusions"
+        force_inclusion_file="../patch/kconfig-force-secure-boot-inclusions"
+
+        # save the new pub key in kernel
+        sed -i "s|^CONFIG_SYSTEM_TRUSTED_KEYS=.*|CONFIG_SYSTEM_TRUSTED_KEYS=\"$SECURE_UPGRADE_DEV_SIGNING_CERT\"|g" $inclusion_file
```
DavidZagury, line 173: include ${inclusion_file} into the common double quotes

patch/kconfig-secure-boot-inclusions
davidpil2002: done

patch/kconfig-secure-boot-inclusions
davidpil2002: done

manage-config
davidpil2002: its working without {} , but I agree to saved the convention

davidpil2002:
> 

Did your suggestions
@saiarcot895  pls can you approve/comment?

davidpil2002 (commented):
move the Secure Boot kernel config to manage-config instead using an standalone bash script

manage-config
```diff
@@ -58,88 +67,113 @@ function get_section_opts(){
...
+}
+
+exclusion_file="../patch/kconfig-exclusions"
+inclusion_file="../patch/kconfig-inclusions"
+force_inclusion_file="../patch/kconfig-force-inclusions"
+ret_process_inc_ex=0
+ret_process_inc_ex=$(process_inclusion_exclusion_files > /dev/null; echo $?)
+
+#  Secure Boot support
+if [ $ret_process_inc_ex -e 0 ]; then
```
saiarcot895, line 165: I think this should be `-eq`?

Makefile
```diff
@@ -123,7 +123,7 @@ $(addprefix $(DEST)/, $(MAIN_TARGET)): $(DEST)/% :
 
 	# Optionally add/remove kernel options
 	if [ -f ../manage-config ]; then
-		../manage-config $(CONFIGURED_ARCH) $(CONFIGURED_PLATFORM)
+		../manage-config $(CONFIGURED_ARCH) $(CONFIGURED_PLATFORM) $(SECURE_UPGRADE_MODE) $(SECURE_UPGRADE_DEV_SIGNING_CERT)
```
saiarcot895, line 126: Should `SECURE_UPGRADE_MODE` be initialized somewhere in this file?

manage-config
```diff
@@ -58,88 +67,113 @@ function get_section_opts(){
...
+    if [ ${SECURE_UPGRADE_MODE} == "dev" -o ${SECURE_UPGRADE_MODE} == "prod" ]; then
+        echo "set kconfig-secure-boot-exclusions & kconfig-secure-boot-inclusions"
+        exclusion_file="../patch/kconfig-secure-boot-exclusions"
+        inclusion_file="../patch/kconfig-secure-boot-inclusions"
+        force_inclusion_file="../patch/kconfig-force-secure-boot-inclusions"
+
+        # save the new pub key in kernel
+        sed -i "s|^CONFIG_SYSTEM_TRUSTED_KEYS=.*|CONFIG_SYSTEM_TRUSTED_KEYS=\"$SECURE_UPGRADE_DEV_SIGNING_CERT\"|g" ${inclusion_file}
```
saiarcot895, line 173: I would recommend adding a check here to make sure `SECURE_UPGRADE_DEV_SIGNING_CERT` is a non-empty string, at the very least.

saiarcot895: Hmm, actually, instead of modifying `patch/kconfig-secure-boot-inclusions`, I would recommend just setting `CONFIG_SYSTEM_TRUSTED_KEYS` separately, since the value of that string will need to be what the user passes in.

So there would be a line here calling `scripts/config --file ${CONFIG_FILE} -k --set-val "CONFIG_SYSTEM_TRUSTED_KEYS" "$SECURE_UPGRADE_DEV_SIGNING_CERT"`

zhangyanzhao:
@davidpil2002  can you please help to address the comments from @saiarcot895 ? After that, @saiarcot895  will do a re-check, then we should be good. Thanks.

manage-config
davidpil2002: done

manage-config
davidpil2002: about the string exiting, I think this condition covers it:
`if [ $# -ge 4 ]; then`
`SECURE_UPGRADE_DEV_SIGNING_CERT=$4`
`fi`
but, I will add a condition to check that the file really exists to be more accurate.
In addition, I will do your suggestion about the CERT assignment by the directly call of `script/config`

Makefile
davidpil2002: its initiated by the slave container so didn't suffer an uninitialized issue, but I will add it anyway, cannot damage the code

manage-config
davidpil2002: done the (1) comment, but about (2), for some reason when using directly `scripts/config` it's not appending the new certificate. 
So, for now, I will remain with my original solution that modifying the default value from the `patch/kconfig-secure-boot-inclusions`  with the dynamic cert input from the user.
To avoid any issues the script is now testing that the certificate file exists so will no be functional issues.

davidpil2002 (commented):
Done all the fixes about supporting secure boot in config kernel.

liat-grozovik:
@saiarcot895 could you please help to re-review and approve?

saiarcot895 (approved):

saiarcot895:
@davidpil2002 can you merge in the changes from the master branch?

saiarcot895:
@lguohan could you re-review?

patch/kconfig-secure-boot-exclusions
```diff
@@ -0,0 +1,21 @@
+[common]
+
+[amd64]
+CONFIG_MODULE_SIG_SHA256
+# For mellanox
```
lguohan, line 5: why these are excluded? can we have more explanation for the justifications?

davidpil2002: SHA256 is excluded because we are using SHA512. Its more secure.
there is more description in the HLD link attached in the description of this PR.
About the lockdown, we have a plan to support it in the feature, for now for Mellanox its disabled.

davidpil2002: > I'm planning to move the SHA512 config to apply to all kernel builds in a future PR; this isn't necessarily secure-boot specific, and is nice to have in general.

Sound good, can we for now save the PR as is, and when you modify the general config you can remove it.

Because we are implicitly signing the kernel modules with sha512 in the sonic-buildimage:
https://github.com/sonic-net/sonic-buildimage/pull/12692/files#diff-de80d4961ffb88d808888c6d160af8717e70ec6c21675b0f5124b0d27db7a166 
So, if the kernel configuration does not match, the image will not boot.

lguohan (approved):

PR #300, opened 2022-11-17
Fix setting a config with an already-existing conflicting value

### LINKED ISSUE #299: Error when modifying kernel flag with kconfig-inclusions files
**Flow:**
Modifying a kernel flag of string type (flag example below) in the filename: sonic-linux-kernel/patch/kconfig-inclusions.
the  expectation is that after the kernel build we should found a kernel .config file including this modification, but its not happened.
The kernel flag in the /linux-5.10.46/debian/build/build_amd64_none_amd64/.config remained like was at before the modification in the kconfig-inclusions.

**kernel flag example:** 
`CONFIG_SYSTEM_TRUSTED_KEYS="debian/certs/new_cert.key`

**How to repro:**
tested on branch: master
1. add a line in kconfig-inclusions file with the kernel flag, i.e

`CONFIG_SYSTEM_TRUSTED_KEYS="debian/certs/new_cert.key`

2. make the kernel
3. read the kernel .config file created in  /linux-5.10.46/debian/build/build_amd64_none_amd64/.config
4. check the value of the keyname `CONFIG_SYSTEM_TRUSTED_KEYS`.

**Note:**
I think the problem is happening just with kernel flags that contain some string default values, like in this case:
`CONFIG_SYSTEM_TRUSTED_KEYS` have a default value of `"debian/certs/debian-uefi-certs.pem`

saiarcot895:
Fix setting a config value in kconfig-inclusions when there's already a conflicting existing value in defconfig.

For example, trying to overwrite `CONFIG_SYSTEM_TRUSTED_KEYS` would have no effect, because there's already a value specified in the file, and the existing value takes precedence.

Also support verifying the value of strings (and not just y/m/n settings). `CONFIG_SYSTEM_TRUSTED_KEYS` takes a string value, and the current code doesn't completely handle strings.

Fixes #299.

### DIFF
--- a/manage-config
+++ b/manage-config
@@ -115,6 +115,8 @@ if [ -e ${exclusion_file} -o -e ${inclusion_file} -o -e ${force_inclusion_file} 
             if [ ! -z "$opt" ] && [[ ! "$opt" =~ ^#.* ]]; then
                 n=${opt%=*}
                 v="${opt#*=}"
+                v="${v/#\"/}"
+                v="${v/%\"/}"
                 s=$(scripts/config --file ${CONFIG_FILE} -k --state $n)
                 if [ ! "$s" = "$v" ]; then
                     ret=2
--- a/manage-config
+++ b/manage-config
@@ -79,7 +79,9 @@ if [ -e ${exclusion_file} -o -e ${inclusion_file} -o -e ${force_inclusion_file} 
         inclusion_opts=$(get_section_opts ${inclusion_file} "common" ${ARCH} ${PLATFORM})
         while read -r opt; do
             if [ ! -z "$opt" ] && [[ ! "$opt" =~ ^#.* ]]; then
-                echo $opt >> ${CONFIG_FILE}
+                n=${opt%=*}
+                v="${opt#*=}"
+                scripts/config --file ${CONFIG_FILE} -k --set-val "$n" "$v"
             fi
         done <<< ${inclusion_opts};
     fi


paulmenzel:
Thank you for the patch. Could you please split

> Also support verifying the value of strings (and not just y/m/n settings).

into a separate patch?

Also, could you amend the commit message and give an example, that didn’t work before, and works now?

saiarcot895:
Commits have been split into two separate commits

saiarcot895:
@davidpil2002 could you confirm that with this fix, you can overwrite `CONFIG_SYSTEM_TRUSTED_KEYS`?

davidpil2002:
> @davidpil2002 could you confirm that with this fix, you can overwrite `CONFIG_SYSTEM_TRUSTED_KEYS`?
Hi,
Its not working, I think is an issue now related to the bracket ""
this is what I did:
I add this configuration in: patch/kconfig-inclusions
`CONFIG_SYSTEM_TRUSTED_KEYS="debian/certs/nv_onyx_key_certificate.pem"
CONFIG_MODULE_SIG_KEY=""
CONFIG_MODULE_SIG_SHA512=y
CONFIG_MODULE_SIG_HASH=sha512`

The result have this error:
`#
# configuration written to .config
#
make[2]: Leaving directory '/sonic/src/sonic-linux-kernel/linux-5.10.140/debian/build/build_amd64_none_amd64'

Checking removed kernel options...
No error

Checking added kernel options...
Option CONFIG_SYSTEM_TRUSTED_KEYS should be set to ["debian/certs/nv_onyx_key_certificate.pem"] instead of [debian/certs/nv_onyx_key_certificate.pem]
Option CONFIG_MODULE_SIG_KEY should be set to [""] instead of []
Option CONFIG_SECURITY_LOCKDOWN_LSM_EARLY should be set to [n] instead of [undef]
Option CONFIG_LOCK_DOWN_KERNEL_FORCE_NONE should be set to [n] instead of [undef]
Option CONFIG_LOCK_DOWN_IN_EFI_SECURE_BOOT should be set to [n] instead of [undef]
`

In addition, if you write the configuration without bracket "", like the follow:
`CONFIG_SYSTEM_TRUSTED_KEYS=debian/certs/nv_onyx_key_certificate.pem`

the result have this error:
`make[2]: Leaving directory '/sonic/src/sonic-linux-kernel/linux-5.10.140/debian/build/build_amd64_none_amd64'

Checking removed kernel options...
No error

Checking added kernel options...
Option CONFIG_SYSTEM_TRUSTED_KEYS should be set to [debian/certs/nv_onyx_key_certificate.pem] instead of [debian/certs/debian-uefi-certs.pem]
Option CONFIG_MODULE_SIG_KEY should be set to [""] instead of []
Option CONFIG_SECURITY_LOCKDOWN_LSM_EARLY should be set to [n] instead of [undef]
Option CONFIG_LOCK_DOWN_KERNEL_FORCE_NONE should be set to [n] instead of [undef]
Option CONFIG_LOCK_DOWN_IN_EFI_SECURE_BOOT should be set to [n] instead of [undef]`

liat-grozovik:
@davidpil2002 kindly reminder to review and confirm so we can merge this one.

davidpil2002:
> @davidpil2002 kindly reminder to review and confirm so we can merge this one.

Hi @liat-grozovik ,

I reviewed/tested locally the community fix, it's solved some config entries, but still had errors and not supported all the kernel configuration that required.
I wrote a comment with all the details 12 days ago.
Thanks
David

saiarcot895:
For `CONFIG_SECURITY_LOCKDOWN_LSM_EARLY`, `CONFIG_LOCK_DOWN_KERNEL_FORCE_NONE`, and `CONFIG_LOCK_DOWN_IN_EFI_SECURE_BOOT`, you'll want to specify these in `patch/kconfig-exclusions` instead. That'll unset these configs.

With them in `patch/kconfig-inclusions`, you're seeing the error because these configs are either set with `y`, or they're unset (which is implied to mean `n`); they're never explicitly set with `n`.

davidpil2002:
> For `CONFIG_SECURITY_LOCKDOWN_LSM_EARLY`, `CONFIG_LOCK_DOWN_KERNEL_FORCE_NONE`, and `CONFIG_LOCK_DOWN_IN_EFI_SECURE_BOOT`, you'll want to specify these in `patch/kconfig-exclusions` instead. That'll unset these configs.
> 
> With them in `patch/kconfig-inclusions`, you're seeing the error because these configs are either set with `y`, or they're unset (which is implied to mean `n`); they're never explicitly set with `n`.


well in general linux kernel config support to set y or n.
But, I tested now like you sad using the sonic kconfig-exclustions, and its working good.
So, for my POV the fix works and you are welcome to merge

davidpil2002:
> > For `CONFIG_SECURITY_LOCKDOWN_LSM_EARLY`, `CONFIG_LOCK_DOWN_KERNEL_FORCE_NONE`, and `CONFIG_LOCK_DOWN_IN_EFI_SECURE_BOOT`, you'll want to specify these in `patch/kconfig-exclusions` instead. That'll unset these configs.
> > With them in `patch/kconfig-inclusions`, you're seeing the error because these configs are either set with `y`, or they're unset (which is implied to mean `n`); they're never explicitly set with `n`.
> 
> well in general Linux kernel config support setting `y` or `n`. But, I tested now like you said using the sonic `patch/kconfig-exclusions`, and it's working well. So, for my POV the fix works and you are welcome to merge
@saiarcot895 , @liat-grozovik

yxieca (approved):

davidpil2002:
@saiarcot895 
I see that the fix its merged, but pls can you update the pointer of sonic-buildimage to include your fix, so I can remove my WA from my code and use your fix

saiarcot895:
@davidpil2002 I can, but for using my fix, you don't need me to update the sonic-buildimage pointer. You should be able to rebase your private branch on top of the current master branch (or do a merge from the current master branch) and test your changes that way.

PR #311, opened 2023-03-17
[marvell-arm64]: Add support for 98DX35xx and 98CX85xx platform

pavannaregundi:
These patches add necessary support for Marvell's Prestera family of switching platforms, namely 98DX35xx and 98CX85xx. 
98DX35xx is internally referred as AC5(AlleyCat5) or AC5X(AlleyCat5X) and 98CX85xx as Falcon.

**Testing:**
Build of target/sonic-marvell-arm64.bin, load of the image on AC5XRD(98DX35xx) and Falcon(98CX85xx) switches, verified that all dockers operational and “show interface status” displays all ports correctly. Also verified operations of components such as management port, i2c, gpio, watchdog.

Build of target/sonic-marvell-armhf.bin, load of the image on M0, verified that all dockers operational and “show interface status” displays all ports correctly.


pavannaregundi:
@Pavan-Nokia please review PR

saiarcot895:
It looks like some of these patches are backports of existing patches in the kernel, whereas others are custom patches. For the backported patches, could you add the original commit hash, either into the patch file or in the PR description? For the custom patches, are there any plans to upstream these patches into the kernel?

Also, I might be missing something here, but what's the reason of having `marvell-arm64` vs `arm64` in the kconfig inclusion/exclusion files?

pavannaregundi:
@saiarcot895 Commit hash of upstreamed patches are mentioned as part of commit description. 
arm64 is ARCH and marvell-arm64 will be PLATFORM for build. So as per manage-config, kconfig inclusions and exclusions can be created with platform section also. Specifying a platform name adds a more specific filter during the build.

pavannaregundi:
@saiarcot895 Regarding custom patches, Marvell will be planning to upstream them in near future.

patch/kconfig-exclusions
```diff
@@ -15,6 +15,10 @@ CONFIG_THERMAL_STATISTICS
 
 [arm64]
 
+[marvell-arm64]
+CONFIG_UBIFS_FS_ZSTD
+CONFIG_CRYPTO_DEV_SAFEXCEL
```
saiarcot895, line 20: Any reason these two kernel configs are being unset here? Is it just that they're not used, or that they break something? If the former, I'd like to keep them enabled.

saiarcot895:
If possible, I'd like to keep the kernel configs the same across different platforms within a given architecture (like how it is for amd64 currently). I do see that currently armhf does have separate configs for armhf and marvell-armhf, but for arm64 currently, with the kernel configs that are being set, they don't seem like they would break any other platform.

patch/kconfig-exclusions
pavannaregundi: @saiarcot895 Below are the reasons for exclusions. Same information is updated in commit summary.

Overcome following dmesg error during boot.
    - UBIFS error: cannot initialize compressor zstd, error -2
        - CONFIG_UBIFS_FS_ZSTD

    - crypto-safexcel f2800000.crypto: HW init failed (-2)
        - CONFIG_CRYPTO_DEV_SAFEXCEL
        - EIP197 cryptographic engine is not used.

pavannaregundi:
> 

I have moved all the configs except CONFIG_MARVELL_AC5_WATCHDOG under arm64. 
CONFIG_MARVELL_AC5_WATCHDOG is retained under 'marvell-arm64' because this config is getting added as part of the specific kernel patch(0015-ac5-marvell-Add-watchdog-support.patch).

radha-danda:
@saiarcot895, @pavannaregundi addressed your comments. Kindly review the PR and share if you have any further comments pls

paulmenzel:
@radha-danda, please refrain from adding such ping comments only a few hours after the last update. It only causes noise.

lguohan:
maybe i am missing something, but i couldn't find the kernel upstream commit id for any of the patch.

patch/series
```diff
@@ -189,8 +189,29 @@ cisco-hwmon-pmbus_core-pec-support-check.patch
 0003-psample-define-the-macro-PSAMPLE_MD_EXTENDED_ATTR.patch
 
 #
-# Marvell platform patches for 4.19
+# Marvell platform patches for 5.10
 armhf_secondary_boot_online.patch
```
lguohan, line 193: has this patch being upstreamed?

pavannaregundi: > has this patch being upstreamed?

Please find upstreamed patches in the commit description and link has been given to them.

pavannaregundi: Snippet from commits, referencing the upstream commit information.
image

pavannaregundi:
@saiarcot895 Can this PR be backported to 202211 sonic-linux-kernel branch?

StormLiangMS:
@Blueve could we have an ADO to track this cherry pick?

StormLiangMS:
@pavannaregundi cherry pick conflict for 202305, could you help to raise separate PR?

@Blueve for vis.

radha-danda:
@pavannaregundi is on PTO till 25th Oct. He raise PR once he is back to work.

PR #313, opened 2023-03-27
Provision the ability to apply the non upstream patches in any order

vivekrnv:
Currently, non upstream patches are applied only after upstream patches. 

**What i Did**

1) Apply a patch to the series files and copy the relevant patches into the patch/ folder
2) Non upstream Patches that reside in the sonic repo will not be saved in a tar file bur rather in a folder pointed out by `EXTERNAL_KERNEL_PATCH_LOC`. The build variable name is also updated to `INCLUDE_EXTERNAL_PATCHES` 
3) `INCLUDE_EXTERNAL_PATCHES` will be a mandatory parameter to include NON_UPSTREAM_PATCHES

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -65,15 +65,14 @@ $(addprefix $(DEST)/, $(MAIN_TARGET)): $(DEST)/% :
 	rm -rf $(NON_UP_DIR)
 	mkdir -p $(NON_UP_DIR)
 
-	if [ ! -z ${EXTERNAL_KERNEL_PATCH_URL} ];  then
-		wget $(EXTERNAL_KERNEL_PATCH_URL) -O patches.tar
-		tar -xf patches.tar -C $(NON_UP_DIR)
-	fi
-
-	# Precedence is given for external URL
-	if [ -z ${EXTERNAL_KERNEL_PATCH_URL} ] && [ x${INCLUDE_EXTERNAL_PATCHES} == xy ]; then
-		if [ -d "$(EXTERNAL_KERNEL_PATCH_LOC)" ]; then
-			cp -r $(EXTERNAL_KERNEL_PATCH_LOC)/* $(NON_UP_DIR)/
+	if [ x${INCLUDE_EXTERNAL_PATCHES} == xy ]; then
+		if [ ! -z ${EXTERNAL_KERNEL_PATCH_URL} ]; then
+			wget $(EXTERNAL_KERNEL_PATCH_URL) -O patches.tar
+			tar -xf patches.tar -C $(NON_UP_DIR)
+		else
+			if [ -d "$(EXTERNAL_KERNEL_PATCH_LOC)" ]; then
+				cp -r $(EXTERNAL_KERNEL_PATCH_LOC)/* $(NON_UP_DIR)/
+			fi
 		fi
 	fi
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -61,6 +61,37 @@ ORIG_FILE_URL = "$(SOURCE_FILE_BASE_URL)/$(ORIG_FILE)"
...
+	fi
+
+	# Precedence is given for external URL
+	if [ -z ${EXTERNAL_KERNEL_PATCH_URL} ] && [ x${INCLUDE_EXTERNAL_PATCHES} == xy ]; then
+		if [ -d "$(EXTERNAL_KERNEL_PATCH_LOC)" ]; then
+			cp -r $(EXTERNAL_KERNEL_PATCH_LOC)/* $(NON_UP_DIR)/
+		fi
+	fi
+
+	if [ -f "$(NON_UP_DIR)/series.patch" ]; then
```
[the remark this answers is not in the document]
vivekrnv: It just moves ahead without applying the non-upstream patches.

Makefile
```diff
@@ -61,6 +61,37 @@ ORIG_FILE_URL = "$(SOURCE_FILE_BASE_URL)/$(ORIG_FILE)"
...
+	rm -rf $(NON_UP_DIR)
+	mkdir -p $(NON_UP_DIR)
+
+	if [ ! -z ${EXTERNAL_KERNEL_PATCH_URL} ];  then
+		wget $(EXTERNAL_KERNEL_PATCH_URL) -O patches.tar
+		tar -xf patches.tar -C $(NON_UP_DIR)
+	fi
+
+	# Precedence is given for external URL
+	if [ -z ${EXTERNAL_KERNEL_PATCH_URL} ] && [ x${INCLUDE_EXTERNAL_PATCHES} == xy ]; then
```
saiarcot895, line 74: I think we had a discussion on the variable naming in #296, but since this is being renamed to `INCLUDE_EXTERNAL_PATCHES`, can `EXTERNAL_KERNEL_PATCH_URL` fall under that, such that if `INCLUDE_EXTERNAL_PATCHES` is set to `n`, then nothing happens, even if `EXTERNAL_KERNEL_PATCH_URL` is specified?

vivekrnv: That means user has to provide both `EXTERNAL_KERNEL_PATCH_URL  `& `INCLUDE_EXTERNAL_PATCHES `, if they wish to use non-upstream patches from an external URL.  Why have two control options?  `INCLUDE_EXTERNAL_PATCHES` is for locally saved non-up patches. atleast that's how i thought of

vivekrnv: @saiarcot895 Do you think it makes more sense to make `INCLUDE_EXTERNAL_PATCHES` a mandatory parameter.

saiarcot895: My thought was that because of the name `INCLUDE_EXTERNAL_PATCHES`, it is a parameter whether to include external patches at all, and that `EXTERNAL_KERNEL_PATCH_URL` is one source of those external patches.

vivekrnv: Hmm, will update it then

saiarcot895 (approved):

vivekrnv:
Request for 202211

dgsudharsan:
@saiarcot895 @StormLiangMS  Can you please add request for 202211 tag?

keboliu:
@StormLiangMS would you please help to cherry-pick?

PR #320, opened 2023-06-02
[202012] Add crypto/zstd.ko module via CONFIG_CRYPTO_ZSTD

Staphylo:
The feature `docker_inram` gets the maximum compression ratio when using the `zstd` algorithm.

On more recent releases sonic releases the crypto/zstd.ko module is built by default.
It is however not the case on sonic-linux-kernel 4.19 which this change addresses.


paulmenzel (commented):
> The feature docker_inram gets the maximum compression ratio when using the zstd algorithm.

I would have thought LZMA would compresses better?

> On more recent releases sonic releases the crypto/zstd.ko module is built by default.  It is however not the case on sonic-linux-kernel 4.19 which this change addresses.

Can you reference a commit, where it was turned on?

On what device did you test this?

paulmenzel:
Thank you for the detailed comment. LZMA (XZ) is apparently not supported. The guess regarding Debian’s configuration sounds reasonable.

It’d be great, if you amended the commit message to add all that information.

saiarcot895 (approved):

PR #321, opened 2023-06-14
arm64: dts: marvell: Add Nokia 7215-IXS-A1 board

Pavan-Nokia:
Create platform specific dts for marvell -arm64 based Nokia 7215-IXS-A1 platform
This dts is derived from Marvell RD-AC5X board.

differences between RD-AC5X and 7215-A1 is only the default LED setting for Mangement port.

### DIFF
--- a/patch/0021-arm64-dts-marvell-Add-Nokia-7215-IXS-A1-board.patch
+++ b/patch/0021-arm64-dts-marvell-Add-Nokia-7215-IXS-A1-board.patch
@@ -1,0 +1,143 @@
+From 79de58e232cf5880a8ff68f15091febfc6c8fb3b Mon Sep 17 00:00:00 2001
+From: Natarajan Subbiramani <natarajan.subbiramani.ext@nokia.com>
+Date: Tue, 13 Jun 2023 17:04:23 +0000
+Subject: [PATCH] arm64: dts: marvell: Add Nokia 7215-IXS-A1 board
+
+This dts is derived from Marvell RD-AC5X board to 
+create platform specific dts for Nokia 7215-IXS-A1 
+
+Signed-off-by: Natarajan Subbiramani <natarajan.subbiramani.ext@nokia.com>
+Tested-by:  Natarajan Subbiramani <natarajan.subbiramani.ext@nokia.com>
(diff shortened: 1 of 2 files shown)


Pavan-Nokia:
@saiarcot895 Could you please help review and merge

paulmenzel:
Please amend the commit message to note the differences to the Marvell RD-AC5X.

Additionally, please post the URL to the patch, that you send (Linux kernel) upstream for review.

saiarcot895:
Are there plans to upstream this?

Pavan-Nokia:
Hi Paul, SaiKrishna,

We did not think we needed to upstream this as it is specific to sonic. can we have this only in sonic?
We are evaluating now to see if we can/should upstream this patch.

Do you think it is better if we move this under platform instead of linux-kernel?

differences between RD-AC5X and 7215-A1 is only the default LED setting for Mangement port.

saiarcot895:
> Do you think it is better if we move this under platform instead of linux-kernel?

I'm fine with having this be in sonic-linux-kernel (it might actually need to be, since this is a device definition file and would need to be part of the kernel itself for proper bootup of the device).

> We did not think we needed to upstream this as it is specific to sonic. can we have this only in sonic?
> We are evaluating now to see if we can/should upstream this patch.

Unless there's a good reason _not_ to upstream a patch (such as SONiC making some change that would break something for non-SONiC use cases), the preference should be to upstream it, especially since this is only a device definition.

saiarcot895 (approved):

Pavan-Nokia:
Hi Saikrishna, Paul,

Thanks for the merge.
We will work on upstreaming the patch and add the link to this PR within the next 2 weeks.

This device was primarily built only for SONiC that is the reason we were thinking of keeping this patch only in Sonic-linux-kernel

Pavan

Pavan-Nokia:
Hi Saikrishna, Paul,

Here is the link to the upstream merge request for this patch 
https://lore.kernel.org/all/20230817155736.1221-1-nataccet@gmail.com/

The community has requested to use latest dt-binding mechanism for LED phy related changes available in upstream master. The support for this mechanism is not yet present in current sonic-linux-kernel. We will create an equivalent PR in the future when support is available.

Pavan

PR #324, opened 2023-07-18
Fix Makefile syntax and provide default value for CONFIGURED_PLATFORM

saiarcot895:
Fix the assignment syntax for SECURE_UPGRADE_SIGNING_CERT. Also, provide a default value for CONFIGURED_PLATFORM, since that's used in the call to `manage-config`.


paulmenzel (approved):
Two commits would be nice, but the diff is good.

lguohan (approved):

PR #325, opened 2023-08-22
[PATCH] net: allow user to set metric on default route learned via Router Advertisement

abdosi:
What I did:
Patch taken from Linux 5.12 to allow user to set metric on default route learned via Router Adv message for IPv6

This PR is to backport the following commits to kernel v5.10:

https://github.com/torvalds/linux/commit/6b2e04bc240fe9be9e690059f710e9f95346d34d

Why I did:

Fixes: https://github.com/sonic-net/sonic-buildimage/issues/16209


abdosi:
@rlhui for viz.

saiarcot895:
Can this patch be applied to the master branch first, and then backported from there?

saiarcot895:
Also, please clean up the title of the PR, and add the upstream commit hash into the patch.

abdosi:
> Can this patch be applied to the master branch first, and then backported from there?

as the master will be moving to new Debian with Linux 6.1 do we still want to take it into master  ?

rlhui:
@abdosi is it correct that this one is closed?

PR #326, opened 2023-08-23
[PATCH] net: allow user to set metric on default route learned via  Router Advertisement

abdosi:
What I did:
Patch taken from Linux 5.12 to allow user to set metric on default route learned via Router Adv message for IPv6

This PR is to backport the following commits to kernel v5.10:

https://github.com/torvalds/linux/commit/6b2e04bc240fe9be9e690059f710e9f95346d34d

ADO: 24888962

Why I did:

Fixes: https://github.com/sonic-net/sonic-buildimage/issues/16209


abdosi:
@saiarcot895 created PR for master

saiarcot895 (approved):

paulmenzel:
What does the line below mean?

> ADO: 24888962

zhenggen-xu:
@yxieca @StormLiangMS Can you confirm if this kernel patch is in 202305 branch?  I only see the config changes in sonic-buildimage 202305 branch but not the Linux kernel patch, that won't be sufficient.  thanks!

PR #329, opened 2023-09-07
Change the system.map file permission only readable by root

xumia:
Change the system.map file permission only readable by root

### DIFF
--- a/patch/0001-Change-the-system.map-file-permission-only-readable-.patch
+++ b/patch/0001-Change-the-system.map-file-permission-only-readable-.patch
@@ -1,0 +1,25 @@
+From 01e598f75f4ab650555b01116ceec4e5c8f2899b Mon Sep 17 00:00:00 2001
+From: xumia <xumia@contoso.com>
+Date: Thu, 7 Sep 2023 02:53:49 +0000
+Subject: [PATCH] Change the system.map file permission only readable by root
+
+---
+ debian/rules.real | 2 +-
+ 1 file changed, 1 insertion(+), 1 deletion(-)
+
+diff --git a/debian/rules.real b/debian/rules.real
+index 3304579ad..908258789 100644
+--- a/debian/rules.real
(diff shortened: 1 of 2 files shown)


saiarcot895:
Looks like this was opened against an older version of the master branch, can you merge/rebase against the latest changes?

saiarcot895:
Never mind, I see you updated it now

saiarcot895 (approved):

paulmenzel:
@saiarcot895, why did you merge a “security patch“, that does not contain any information at all about the motivation? Why are Debian’s defaults bad?

@saiarcot895, please revert, and @xumia, please resubmit with proper description.

saiarcot895:
@paulmenzel there are some details in sonic-net/sonic-buildimage#15983.

The short version of it is that as per OpenSCAP, the [System.map file must be readable only by root](http://static.open-scap.org/ssg-guides/ssg-debian11-guide-standard.html#xccdf_org.ssgproject.content_rule_file_permissions_systemmap). This is despite the fact that Debian already ships a fake System.map file (since there's almost never a need to use the actual contents of the file at runtime). The options to meet OpenSCAP's requirements are either to make it readable only by root, or to remove it (both of which can be done either at package installation time or during the package build time). Since it would be better to have it be consistently done at package build time (so that manual changes to files from packages are not needed), I recommended having the change be done in this repo via a patch instead.

To be clear, it's not that Debian's defaults are bad/insecure; it's just that some security projects/audits have different requirements that need to be met.

saiarcot895:
@xumia This patch actually needs to go into the `patches/preconfig` directory to have any effect. Could you make another PR to fix this?

xumia:
> @xumia This patch actually needs to go into the `patches/preconfig` directory to have any effect. Could you make another PR to fix this?

@saiarcot895 , thanks, I have sent another PR to fix it, https://github.com/sonic-net/sonic-linux-kernel/pull/331

PR #334, opened 2023-09-26
arm64: ac5: Fix watchdog timeleft

pavannaregundi:
Watchdog timeleft returns unexpected huge value for first read, followed by correct expected values. Unexpected huge value reappears after few minutes. Issue is resolved by updating the existing hardware workaround to read WCV_HI along with WCV_LO.

Changes also needs to be cherry picked to 202211 branch.


saiarcot895 (approved):

StormLiangMS:
@pavannaregundi have you test with 202305?

pavannaregundi:
@StormLiangMS I have tested this on master. Kernel and this driver same for 202305. This may not need separate testing.

PR #336, opened 2023-09-28
Add nvidia arm section and an ability to patch kconfig-inc and fix manage-config

vivekrnv:
1) Add non-amd64 nvidia section
2) Add an ability to patch kconfig-inclusions file when applying non-upstream patches
3) Fix two bugs in manage-config script 
      - process_inclusion_exclusion_files method is not shown in the log file since it is redirected to /dev/null
      - platform-arch sections in kconfig-inclusions & kconfig-exclusions are nto parsed today by the manage config script


dgsudharsan (approved):

paulmenzel:
`\Add nvidia arm markers and add an ability to patch kconfig-inc` has a stray leading `\`.

Could you please elaborate more, and add commit messages? Markers seem to make the build system even more complicated.

vivekrnv:
> `\Add nvidia arm markers and add an ability to patch kconfig-inc` has a stray leading `\`.
> 
> Could you please elaborate more, and add commit messages? Markers seem to make the build system even more complicated.

We have automatic infrastructure which depend on markers to effectively make changes to these files. They are just comments and doesn't affect the build process. I've mentioned other changes made in the PR description.

Makefile
```diff
@@ -77,19 +79,25 @@ $(addprefix $(DEST)/, $(MAIN_TARGET)): $(DEST)/% :
...
+		git stash -- $(SERIES)
+		git apply $(NON_UP_DIR)/$(SERIES).patch
+	fi
+
+	if [ -f "$(NON_UP_DIR)/$(KCFG_INCL).patch" ]; then
+		echo "Patch the $(KCFG_INCL) file"
+		cat $(NON_UP_DIR)/$(KCFG_INCL).patch
+		git stash -- $(KCFG_INCL)
+		git apply $(NON_UP_DIR)/$(KCFG_INCL).patch
+	fi
```
saiarcot895, line 95: Why are two separate patches needed here? A single patch should be able to handle everything.

vivekrnv: Thanks for pointing out, will update

patch/kconfig-exclusions
```diff
@@ -6,14 +6,20 @@ CONFIG_CGROUP_NET_CLASSID
 CONFIG_NET_CLS_CGROUP
 CONFIG_NETFILTER_XT_MATCH_CGROUP
 CONFIG_CGROUP_NET_PRIO
+###-> mellanox_common-start
+###-> mellanox_common-end
```
saiarcot895, line 10: Are these tags really necessary? This file is (almost) an INI file and should be easy for a python script to parse the sections in this file and append/modify as necessary.

vivekrnv: I wanted to keep the implementation in sync with kconfig-inclusions. Besides, adding [mellanox], [mellanox-arm64] etc would make the kernel platform specific and was not recommended here https://github.com/sonic-net/sonic-linux-kernel/pull/309#issuecomment-1470134897

saiarcot895: I meant for both kconfig-exclusions and kconfig-inclusions (and kconfig-force-inclusions, although I need to see if that's still actually needed or not). For instance, the following bit of Python 3 code can parse both files, allows editing the config settings, and then writing them back:

```
>>> import configparser
>>> import sys
>>> configParser = configparser.ConfigParser(allow_no_value=True, strict=False)
>>> configParser.optionxform = str
>>> configParser.read("patch/kconfig-inclusions")
['patch/kconfig-inclusions']
>>> configParser.write(sys.stdout)
[common]
CONFIG_LOG_BUF_SHIFT = 20

[amd64]
CONFIG_I2C_MUX_PCA9541 = m
CONFIG_SENSORS_MAX6697 = m
CONFIG_PCI_REASSIGN_PREF_MEM = y
...
```

`strict=False` is needed because there are currently options specified twice, which should ideally be cleaned up. Fortunately, those duplicates are setting non-conflicting values. Comments aren't preserved though, making the diff size larger than the minimum, but it still functionally works.

I can understand it's needed for the `patch/series` file, because order matters there more so than the kernel configs, but I don't think the kconfig files need the tags.

vivekrnv: It is certainly doable (ignoring the diff) and is a better approach. it also would need a fix in manage-config (it currently doesn't read platform-arch flags). 

However, it goes against your argument https://github.com/sonic-net/sonic-linux-kernel/pull/309#issuecomment-1470134897 that its preferable to have a platform agnostic kernel. What would you recommend considering that?

saiarcot895: For amd64, the kernel should still be platform-agnostic. For armhf and arm64, I _think_ it would be a bit difficult to have a platform-agnostic kernel (I could be wrong here; my knowledge of armhf and arm64 is limited compared to amd64), so a platform-specific kernel is fine for now.

For manage-config, I think it currently considers arch, and then platform (separately), right?

vivekrnv: the script only fetches, common, platform & arch as of today and applies only those. Not [platform-arch] https://github.com/sonic-net/sonic-linux-kernel/blob/1398567e255474c06de6fa93d5ca17ae7d948f6b/manage-config#L143

So, you'd suggestion would be to keep the existing "mellanox-start/mellanox-end" markers (i.e. for amd64) and for arm64/armhf you'd recommend using the ini approach?

vivekrnv: As discussed online, the amd64 kernel will have same kconfig across all the platforms and so markers are used. 

arm64 can be platform specific and thus a new section [mellanox-arm64] is added.

saiarcot895 (approved):

saiarcot895:
Requesting backport for this since this is an infra-level change requested by Mellanox, and it doesn't affect functionality of the compiled kernel.

vivekrnv:
Backport for 2305 https://github.com/sonic-net/sonic-linux-kernel/pull/342

StormLiangMS:
@saiarcot895 to support cherry pick to old ones, could you have an ADO and update in this PR?

PR #365, opened 2023-11-24
[202305] [kconfig] Set default SATA Link Power Management policy

volodymyrsamotiy:
Between kernel 4.9 and kernel 5.10 new kconfig was introduced - CONFIG_SATA_MOBILE_LPM_POLICY.
It affects the power policy of the SSD.
In 201911 power policy was "max_performance" and it was changed by Debian 11 kernel 5.10 to "med_power_with_dipm" which allows SSD to go to sleep mode.

This change simply configures "CONFIG_SATA_MOBILE_LPM_POLICY" to "1" which means "max_performance".

Power policy can be checked by running "cat /sys/class/scsi_host/host0/link_power_management_policy"


dgsudharsan:
@saiarcot895 @prgeor Can you please review and signoff?

saiarcot895 (approved):

PR #368, opened 2023-12-15
Change the system.map file permission only readable by root

xumia:
## Why I did it
It is to fix the issue that  local System.map file (if exists) is readable only by root, see https://static.open-scap.org/ssg-guides/ssg-debian11-guide-standard.html#xccdf_org.ssgproject.content_rule_file_permissions_systemmap.

The current status:
```
admin@test4:~$ ls -l /usr/lib/debug/boot/System.map-6.1.0-11-2-amd64
-rw-r--r-- 1 root root 3905746 Aug  8 11:18 /usr/lib/debug/boot/System.map-6.1.0-11-2-amd64
admin@test4:~$ head /usr/lib/debug/boot/System.map-6.1.0-11-2-amd64
0000000000000000 D __per_cpu_start
0000000000000000 D fixed_percpu_data
00000000000001ea A kexec_control_code_size
0000000000001000 D cpu_debug_store
0000000000002000 D irq_stack_backing_store
0000000000006000 D cpu_tss_rw
000000000000b000 D gdt_page
000000000000c000 d exception_stacks
admin@test4:~$ ls /boot/System.map-6.1.0-11-2-amd64 -l
-rw-r--r-- 1 root root 83 Dec  6 12:14 /boot/System.map-6.1.0-11-2-amd64
admin@test4:~$ 
```

We already have a fix: https://github.com/sonic-net/sonic-linux-kernel/pull/329, but it was not used for upgrading to bookworm code conflict.

## How I did it
The issue only in debian package, it is not caused by the Linux Kernel.
Change the rules.real build script to set the permission to 600.

It fixes the Symtem.map permission issue in both /boot folder and in /usr/lib/debug/boot/.


xumia:
@saiarcot895 , could you please help review it? Thanks.

saiarcot895 (approved):

PR #370, opened 2023-12-16
Dynamic write timeout support for optoe driver

mihirpat1:
The optoe kernel driver currently has a fixed 25ms write timeout value. We need to provide an interface to support overwriting the write_timeout value from userspace.
The overwriting of the write timeout value is needed for some non-standard optics which are not compliant with the timeout value in spec and will help in resolving issues related to EEPROM access.
The overwriting of the write_timeout value can be done based on platform or vendor of the transceiver.

Following sysfs will be created with this change
```
/sys/bus/i2c/devices/<bus_id>-0050/write_timeout
```

Testing
```
root@sonic:/sys/bus/i2c/devices/31-0050# ls
dev_class  driver  eeprom  modalias  name  port_name  power  subsystem  uevent  write_max 
 write_timeout
root@sonic:/sys/bus/i2c/devices/31-0050#

root@sonic:/sys/bus/i2c/devices/35-0050# echo 500 > write_timeout 
root@sonic:/sys/bus/i2c/devices/35-0050# cat write_timeout 
500
root@sonic:/sys/bus/i2c/devices/35-0050# echo 0 > write_timeout 
root@sonic:/sys/bus/i2c/devices/35-0050# cat write_timeout 
0
root@sonic:/sys/bus/i2c/devices/35-0050# echo 25 > write_timeout 
root@sonic:/sys/bus/i2c/devices/35-0050# cat write_timeout 
25
root@sonic:/sys/bus/i2c/devices/35-0050# 

Invalid testcases
root@sonic:/sys/bus/i2c/devices/35-0050# echo 501 > write_timeout 
bash: echo: write error: Invalid argument
root@sonic:/sys/bus/i2c/devices/35-0050# echo -1 > write_timeout 
bash: echo: write error: Invalid argument
```

ADO - 26145450


prgeor:
@mihirpat1 please specify the real need. i.e we sometime need to support non-standard optics not compliant to the spec

prgeor (dismissed):

mihirpat1:
> @mihirpat1 please specify the real need. i.e we sometime need to support non-standard optics not compliant to the spec

@prgeor - Updated the description now.

patch/series
```diff
@@ -32,6 +32,7 @@ driver-support-optoe-write-max.patch
 driver-support-optoe-twoaddr-a2h-access.patch
 driver-support-optoe-oneaddr-pageable.patch
 driver-support-optoe-update-to-linux-6.1.patch
+driver-support-dynamic-write-timeout.patch
```
saiarcot895, line 35: Can you add optoe to the name for consistency purposes with the other optoe patches?

saiarcot895:
@donboll FYI

patch/driver-support-dynamic-write-timeout.patch
```diff
@@ -0,0 +1,126 @@
...
++
++static ssize_t set_dev_write_timeout_size(struct device *dev,
++			struct device_attribute *attr,
++			const char *buf, size_t count)
++{
++	struct i2c_client *client = to_i2c_client(dev);
++	struct optoe_data *optoe = i2c_get_clientdata(client);
++	int write_timeout_size;
++
++	if (kstrtouint(buf, 0, &write_timeout_size) != 0 ||
```
saiarcot895, line 85: The third argument to `kstrtouint` takes in an `unsigned int *`, but you're passing in an `int *`, please fix this.

patch/driver-support-dynamic-write-timeout.patch
```diff
@@ -0,0 +1,126 @@
...
++		write_timeout_size < 0)
++		return -EINVAL;
++
++	mutex_lock(&optoe->lock);
++	optoe->write_timeout = write_timeout_size;
```
saiarcot895, line 90: Should there be an upper limit set here? I'm wondering if a high timeout is set, and the write takes that long, then either some other kernel process or a userspace process will be stuck for that long.

patch/series
mihirpat1: I have renamed the patch now.

patch/driver-support-dynamic-write-timeout.patch
mihirpat1: Fixed now. Thanks!

patch/driver-support-dynamic-write-timeout.patch
mihirpat1: I have now added a 500ms max timeout restriction now.

patch/driver-support-optoe-dynamic-write-timeout.patch
```diff
@@ -0,0 +1,129 @@
...
++static ssize_t set_dev_write_timeout_size(struct device *dev,
++			struct device_attribute *attr,
++			const char *buf, size_t count)
++{
++	struct i2c_client *client = to_i2c_client(dev);
++	struct optoe_data *optoe = i2c_get_clientdata(client);
++	unsigned int write_timeout_size;
++
++	if (kstrtouint(buf, 0, &write_timeout_size) != 0 ||
++		write_timeout_size < 0 || write_timeout_size > OPTOE_MAX_SUPPORTED_WRITE_TIMEOUT)
```
saiarcot895, line 89: Considering `write_timeout_size` is an unsigned int, `write_timeout_size < 0` will (likely) never be true.

As an alternative, is it better to have a required minimum of 5ms or 20ms?

mihirpat1: Agreed that `write_timeout_size < 0` will likely not be true.
I have currently kept 0ms as minimum timeout value since didn't want to have a pre-defined minimum value which can potentially prevent future wait_timeout optimization. This would be the case if a transceiver in future supports wait_timeout less than pre-defined minimum timeout (although less likely)

saiarcot895 (approved):

PR #375, opened 2024-01-10
Add Kconfig diff generation script between different kernel versions

vivekrnv:
### Why I did it
During Kernel Upgrades, there is a need to generate and document the Kernel Config diff. 

### What I did it
Integrate a kconfig diff generation script into the Azure pipelines for _Azure.sonic-linux-kernel_ Job. Currently the reference buildid is harcoded to the last run where the kernel was 5.10. 

The job would generate and add the file named `kconfig-diff-{platform}-{arch}.rst` into the artifact.

During every kernel upgrade (major & minor), the maintainer of sonic-linux-kernel should update the Wiki to reflect this  diff.   [Wiki_Diff](https://github.com/vivekrnv/sonic-linux-kernel/wiki/KConfig-Difference:-5.10.0%E2%80%9023%E2%80%902%E2%80%90amd64-&-6.1.0%E2%80%9011%E2%80%902%E2%80%90amd64)

Note: The output of the rst file can be copied directly into the text section.

### TODO
Once the wiki is edited for the current kernel version, the reference artifact can be set to the last successful run of the corresponding branch.


kcfg-diff.py
```diff
@@ -0,0 +1,144 @@
...
+                new_val += lis[i][j*width:(j+1)*width]
+                new_val += "\n"
+            else:
+                new_val += lis[i][j*width:]
+        lis[i] = new_val
+
+def format_diff_table(additions, modifications, deletions):
+    max_len = max(len(additions), len(deletions))
+    additions += [''] * (max_len - len(additions))
+    deletions += [''] * (max_len - len(deletions))
```
saiarcot895, line 74: Can you make the additions, deletions, and modifications be sorted?

saiarcot895: Actually, never mind, it looks like it is in the Kconfig order.

kcfg-diff.py
saiarcot895, line 1: You could probably move this into the `.azure-pipelines` folder, since it's meant to be used within the pipeline runs only.

vivekrnv: Resolved

dgsudharsan (approved):

saiarcot895 (approved):

PR #377, opened 2024-02-01
[Marvell-arm64] Fix kexec kernel panic in mvpp2

pavannaregundi:
This change fixes the kernel panic during warm-restart kexec reload.

upstream commit in Linux netdev: https://git.kernel.org/pub/scm/linux/kernel/git/netdev/net.git/commit/drivers/net/ethernet/marvell?id=9f538b415db862e74b8c5d3abbccfc1b2b6caa38


pavannaregundi:
@saiarcot895 Can you please review this.

saiarcot895 (approved):

PR #378, opened 2024-03-04
arm64: dts: marvell: Add DTS for 7215-IXS-A1 board

Pavan-Nokia:
Create platform specific DTS for Marvell-arm64 based Nokia 7215-IXS-A1 platform.

7215-IXS-A1 is an aggregation switch based on Marvell AlleyCat5X. This DTS is extracted from Marvell cn9130-crb (reference board) and removed unused nodes along with platform specific changes.

Upstream Kernel PR
https://lore.kernel.org/linux-arm-kernel/CAJOx763HwgK-EEt-AkVqtoObhx_VGokj-Mn0yhS7Sno0uDOZfw@mail.gmail.com/T/#m3d1fc5bce0dcb5ec001f8496a207ccb485903507


Pavan-Nokia:
@saiarcot895 Please help review and merge

patch/0011-arm64-dts-marvell-Add-Nokia-7215-IXS-A1-board.patch
```diff
@@ -137,6 +257,8 @@ index 49f274add..310b57e47 100644
  dtb-$(CONFIG_ARCH_MVEBU) += ac5-98dx35xx-rd.dtb
  dtb-$(CONFIG_ARCH_MVEBU) += armada-7020-comexpress.dtb
 +dtb-$(CONFIG_ARCH_MVEBU) += 7215-ixs-a1.dtb
+
+base-commit: 6b6f1082cb46d72823b7ea99c058c601668ba1d3
```
saiarcot895, line 261: Is this line meant to be part of the patch description, or the file changes? If it's part of the description, can this be moved to the top, below the existing description?"

Pavan-Nokia: This is not part of the file changes. This is the upstream base commit ID intended for the upstream maintainers review. This is autogenerated when git-format patch executed on upstream repo.
This has no effect on SONiC.

Do you recommend removing this line from SONiC patch?

saiarcot895 (approved):
Regarding the comments in the kernel mail thread, I concur with what Andrew has said there. The patch that is submitted to mainline should be using mainline's features/APIs. For the backport to 5.10, that patch can be modified to fit whatever is available in 5.10.x.

Because SONiC's kernel is based on Debian's kernel, we will always be behind the mainline, so there will always be some level of backporting that would need to be done.

Pavan-Nokia:
@saiarcot895  Understood and agree. The comment you are referring to was from our previous version of the DTS that we tried to upstream.
We have modified the DTS to address most of Andrew's comments in this version of the DTS as much as our platform supports. 
Thank you for help with review and merging

PR #380, opened 2024-03-05
[nvidia-bluefield] Add config inclusions for Nvidia Bluefield-3 DPU

oleksandrivantsiv:
Nvidia BF-3 Data Processing Unit is a 400 gigabits per second (Gb/s) infrastructure compute platform with
line-rate processing of software-defined networking. Nvidia BF-3 DPU can be used as a standalone appliance
as well as one of the modules in the Smart Switch device.


vivekrnv (approved):

patch/kconfig-secure-boot-inclusions
```diff
@@ -11,7 +11,7 @@ CONFIG_MODULE_SIG_FORCE=y
 CONFIG_SYSTEM_TRUSTED_KEYS="debian/certs/debian-uefi-certs.pem"
 CONFIG_MODULE_SIG_HASH="sha512"
 CONFIG_MODULE_SIG_SHA512=y
-CONFIG_KEXEC_SIG_FORCE=y
+CONFIG_KEXEC_SIG=y
```
paulmenzel, line 14: Please make this a separate commit.

paulmenzel (changes requested):
Thank you for your patch. Some small nits:

1.  The [official spelling seems to be *BlueField*](https://www.nvidia.com/en-us/networking/products/data-processing-unit/).
2.  Please split the commit into two.
3.  Please amend the commit message to remove the trailing dot from the commit message summary/title, and to give a little more background about the system. For example, would BlueField-2 now also be supported?

patch/kconfig-secure-boot-inclusions
oleksandrivantsiv: I moved those changes into a separate PR: https://github.com/sonic-net/sonic-linux-kernel/pull/381

oleksandrivantsiv:
> Thank you for your patch. Some small nits:
> 
> 1. The [official spelling seems to be _BlueField_](https://www.nvidia.com/en-us/networking/products/data-processing-unit/).
> 2. Please split the commit into two.
> 3. Please amend the commit message to remove the trailing dot from the commit message summary/title, and to give a little more background about the system. For example, would BlueField-2 now also be supported?

Thank you Paul for your feedback. 
1. Both `Bluefield` and `BlueField` are correct spellings as well as BF-3. I prefer `Bluefield`
2. The default merge strategy in the repository is "squash". Even if I split the changes into two commits they will be squashed together during PR merge. Instead, I split the changes into separate PR: https://github.com/sonic-net/sonic-linux-kernel/pull/381
3. Only BF-3 is supported. We have no plans to add BF-2 support.

patch/kconfig-inclusions
```diff
@@ -200,4 +200,15 @@ CONFIG_I2C_MUX_GPIO=m
...
+[nvidia-bluefield]
+CONFIG_MMC_DW=y
+CONFIG_MMC_DW_PLTFM=y
+CONFIG_MMC_DW_BLUEFIELD=y
+CONFIG_ARM64=y
+CONFIG_ACPI=y
+CONFIG_I2C=y
+CONFIG_I2C_SLAVE=y
+CONFIG_ARCH_SUPPORTS_ACPI=y
+CONFIG_EFIVAR_FS=y
```
saiarcot895, line 212: These lines are, I think, already set by other options or default values. Can they be removed?

vivekrnv: I think none of these are common or repeated

saiarcot895: They would generally apply to all ARM64 platforms. `CONFIG_ARM64`, for example, would also be needed by marvell-arm64.

vivekrnv: Okay, Makes sense to move CONFIG_ARM64 into the common arm64 section. 

But for others we are not certain that they can be safely moved to common section. If you think so, we can update.

oleksandrivantsiv: @saiarcot895 I removed all options enabled by default. Please check

saiarcot895 (approved):

oleksandrivantsiv:
@saiarcot895 can we merge this PR?

PR #381, opened 2024-03-05
Fix the issue with signed kernel compilation for ARM64 architecture

oleksandrivantsiv:
CONFIG_KEXEC_SIG_FORCE is not available for ARM64 architecture. CONFIG_KEXEC_SIG should be used instead.


paulmenzel (approved):

saiarcot895:
Ah, I was trying to figure out why the PR checkers didn't fail, but we don't have a PR check for secure boot yet.

saiarcot895 (approved):

PR #387, opened 2024-03-21
Add i2c imc bus driver for support Broadwell memory sensors

jostar-yang:
Add i2c imc bus driver for support Broadwell memory sensors


saiarcot895:
Can you add more details about what platforms need this?

This looks like it is from https://lore.kernel.org/all/1582498270-50674-1-git-send-email-schaecsn@gmx.net/. If so, could you add that into the description or the PR file?

jostar-yang:
The driver ,i2c-imc.c, support ddr sensors for Broadwell CPU machine. It is i2c bus driver. When probe i2c-imc.c, jc42 will be probed ,too.  Then, we will see 0x18 and 0x1a i2c device address are created. Below Accton platform CPU is Broadwell , 
AS9716
AS9726
AS7726
AS7326
Test log as beow,
root@sonic:/sys/bus/i2c/drivers/jc42/57-0018/hwmon/hwmon46# cat temp1_input
27000
root@sonic:/sys/bus/i2c/drivers/jc42# cd 57-001a/hwmon/hwmon47/
root@sonic:/sys/bus/i2c/drivers/jc42/57-001a/hwmon/hwmon47# cat temp1_input
26125 

1. The code is copied from ONL. But compile fail on kernel 6.x. So I change from "i2c_new_device()" to"i2c_new_client_device".
2. Add I2C_FUNC_SMBUS_WORD_DATA to pass regmap check i2c_func. Otherwise, jc42 will probe fail.

lguohan:
if this is the original upgrade patch, @jostar-yang , can you use the original patch?

lguohan:
please follow guideline here.

Existing kernel modules need to be patched and those patches are available in upstream.
Please include the original upstream commit id and message in the patch. This allows the maintainer to remove upstream patches during the kernel upgrade.

PR #401, opened 2024-06-13
Fix optoe's write_max when adapter supports i2c functions

prgeor:
Optoe driver with write-max > 1 when using the i2c message buffer was allocating only 3(1+2) bytes  which means while copying the user data for i2c adapters supporting the native i2c functions will overwrite the message buffer size. This will not happen for i2c adapters that does not support native i2c functions like pure SMBUS controller.

### DIFF
--- a/patch/driver-support-optoe-write-max_fix.patch
+++ b/patch/driver-support-optoe-write-max_fix.patch
@@ -1,0 +1,35 @@
+From f1250392a9dff619b147d8aee0c73c0f6d550713 Mon Sep 17 00:00:00 2001
+From: Prince George <prgeor@microsoft.com>
+Date: Thu, 13 Jun 2024 14:18:22 +0000
+Subject: [PATCH] Fix optoe's write_max when using native i2c driver
+
+Signed-off-by: Prince George <prgeor@microsoft.com>
+---
+ drivers/misc/eeprom/optoe.c | 4 ++--
+ 1 file changed, 2 insertions(+), 2 deletions(-)
+
+diff --git a/drivers/misc/eeprom/optoe.c b/drivers/misc/eeprom/optoe.c
+index 5a2e71be5..1efd18040 100644
+--- a/drivers/misc/eeprom/optoe.c
++++ b/drivers/misc/eeprom/optoe.c
+@@ -1114,7 +1114,7 @@ static int optoe_probe(struct i2c_client *client,
+ 		goto exit;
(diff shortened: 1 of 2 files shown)


billie-alsup (approved):
This certainly fixes the immediate problem of kernel corruption, however it may have implications for low memory devices with high transceiver counts.  The writebuf is completely unused in the use_smbus != 0 case, so an optimization might be to handle that specific case.  For the pure i2c controllers using i2c_transfer, I had also considered reallocating in the sysfs write function, however that gets tricky with interactions with optoe_eeprom_write (as we don't want to hold a lock across the timeout retry loop).  Finally, another option might be to have a stack based buffer in optoe_eeprom_write for small transfers, and dynamcally allocate/free within optoe_eeprom_write for large transfers (and avoid any allocation at all in the probe function).  But again, this solution solves the immediate problem of kernel corruption, at the expense of more memory usage.

prgeor:
> This certainly fixes the immediate problem of kernel corruption, however it may have implications for low memory devices with high transceiver counts. The writebuf is completely unused in the use_smbus != 0 case, so an optimization might be to handle that specific case. For the pure i2c controllers using i2c_transfer, I had also considered reallocating in the sysfs write function, however that gets tricky with interactions with optoe_eeprom_write (as we don't want to hold a lock across the timeout retry loop). Finally, another option might be to have a stack based buffer in optoe_eeprom_write for small transfers, and dynamcally allocate/free within optoe_eeprom_write for large transfers (and avoid any allocation at all in the probe function). But again, this solution solves the immediate problem of kernel corruption, at the expense of more memory usage.

@billie-alsup I had thought about this and I would prefer to keep the memory allocation once, instead of resizing the memory buffer size to the optoe's write data size causing fragmentation at runtime. I am doing a trade-off between memory (128 bytes * 128 physical ports) = 16KB waste vs memory fragmentation and simplicity that comes with static allocation.

prgeor:
@mihirpat1 fyi

liushilongbuaa:
reopen this PR to run pipeline.

lguohan:
@prgeor , can you open pr on 202311/202405 and master for this bug fix?

AnoopKamath:
Hi @prgeor : is this change available on 202311/202405 and master?

mihirpat1:
202405 PR link
https://github.com/sonic-net/sonic-linux-kernel/pull/410

PR #407, opened 2024-06-18
Fix optoe's write_max when using native i2c driver

prgeor:
Optoe driver with write-max > 1 when using the i2c message buffer was allocating only 3(1+2) bytes which means while copying the user data for i2c adapters supporting the native i2c functions will overwrite the message buffer size. This will not happen for i2c adapters that does not support native i2c functions like pure SMBUS controller.

MSFT ADO - 28209032


prgeor:
@bingwang-ms please include this fix in 202405

prgeor:
@StormLiangMS please include this fix in 202311

prgeor:
@lguohan please review/merge. thanks.

patch/driver-support-optoe-write-max_fix.patch
```diff
@@ -0,0 +1,34 @@
...
+diff --git a/drivers/misc/eeprom/optoe.c b/drivers/misc/eeprom/optoe.c
+index 5a2e71be5..1efd18040 100644
+--- a/drivers/misc/eeprom/optoe.c
++++ b/drivers/misc/eeprom/optoe.c
+@@ -1114,7 +1114,7 @@ static int optoe_probe(struct i2c_client *client,
+ 		goto exit;
+ 	}
+ 
+-	dev_dbg(&client->dev, "dev_class: %d\n", optoe->dev_class);
++	dev_dbg(&client->dev, "dev_class: %d use_smbus=%d\n", optoe->dev_class, use_smbus);
```
saiarcot895, line 20: Is there a reason to having `use_smbus` be logged?

prgeor: @prgeor yes the crash is seen on platform using smbus driver so will be good to know the smbus usage in the debug log

saiarcot895 (approved):

r12f:
thanks Mihir and this one is already merged with 202411 -> 202412 code sync: https://github.com/Azure/sonic-linux-kernel.msft/commit/f3f9a37715f8c2e4ed9821547585c3bbb2612619

PR #408, opened 2024-06-18
[202305][Mellanox]Integrate HW-MGMT 7.0030.2300 Changes

dprital:
**Why I did it**
Integrate HW-MGMT 7.0030.2300 Changes

 ## Patch List
* 0329-platform-mellanox-indicate-deferred-I2C-bus-creation.patch :


keboliu (approved):

saiarcot895 (approved):

patch/kconfig-inclusions
```diff
@@ -83,7 +81,6 @@ CONFIG_MLXREG_IO=m
 CONFIG_MLX_WDT=m
 CONFIG_MLXREG_LC=m
 CONFIG_SENSORS_MLXREG_FAN=m
-CONFIG_GPIO_GENERIC=m
```
lguohan, line 86: this seems generic module, why are we remove GPIO_GENERIC AND REGMAP_I2C?

lguohan: will these removal affect other platforms? @saiarcot895 ?

saiarcot895: In the CI builds, both `CONFIG_GPIO_GENERIC` and `CONFIG_REGMAP_I2C` are still enabled, and set to `m`, likely due to other dependencies, so removing those two configs in this PR has no net impact.

vivekrnv: Both the Kconfig's are removed since they are redundant as saikrishna already pointed out. No harm in removing them from here.

lguohan: thanks for confirmation.

PR #411, opened 2024-07-02
[Micas] Kernel configuration is enabled to support mdio-gpio and spi-…

philo-micas:
Help to enable below MACRO:
MDIO-GPIO and MDIO-BITBANG are implementations that simulate MDIO controllers using GPIO pins.
SPI-GPIO is an implementation that simulates SPI controllers using GPIO pins.


philo-micas:
@paulmenzel @lguohan Help arrange review for this PR, thanks!

saiarcot895 (approved):

PR #413, opened 2024-07-02
Fix os crash caused by optoe when class switch

philo-micas:
Why：
When the device is initially set to optoe2, there is an option ->client [1]. If echo 3>dev_class is manually set at this time, as for optoe3, option ->client [1] is meaningless but the content is still the address data of optoe2, accessing an illegal address will occur, causing the device to crash.

How:
Set option->client[1] to NULL when unregistering device

Closes：https://github.com/sonic-net/sonic-linux-kernel/issues/412


paulmenzel:
> Issue link：https://github.com/sonic-net/sonic-linux-kernel/issues/412

Please use:

> Closes: https://github.com/sonic-net/sonic-linux-kernel/issues/412

paulmenzel:
Could you please add the description to the commit message (`git commit --amend && git push -f`).

philo-micas:
> Could you please add the description to the commit message (`git commit --amend && git push -f`).

Ok,thanks

philo-micas:
As mentioned in issue:https://github.com/sonic-net/sonic-buildimage/issues/19525

patch/0001-fix-os-crash-caused-by-optoe-when-class-switch.patch
```diff
@@ -0,0 +1,36 @@
...
++			optoe->client[1] = NULL;
++		}
+ 		optoe->bin.size = ONE_ADDR_EEPROM_SIZE;
+ 		optoe->num_addresses = 1;
+ 	}
+@@ -1226,6 +1228,7 @@ static int optoe_probe(struct i2c_client *client,
+ 	if (num_addresses == 2) {
+ 		if (optoe->client[1])
+ 			i2c_unregister_device(optoe->client[1]);
++			optoe->client[1] = NULL;
```
saiarcot895, line 30: Missing braces around the if-block.

philo-micas: @saiarcot895  Has updated. Please help to approve and merge, thanks!

saiarcot895 (approved):

PR #425, opened 2024-08-19
Set CONFIG_TCG_TPM to m and disable CONFIG_IMA

Junchao-Mellanox:
**Why I did this?**

There is an error kernel log:

```
ERR kernel: [    3.434153] tpm tpm0: [Firmware Bug]: TPM interrupt not working, polling instead
```

To suppress this error log, we need set TPM module to m instead of y so that we can use kernel option to disable TPM module if need.

**What I did?**

Currently, TPM is disabled to y in sonic kernel 6.10. In order to set it to m, we need to:

1. Set CONFIG_IMA to n
2. Set CONFIG_TCG_TPM to m

**How I test this?**

1. Manual check that the error log is disappeared
2. Run sonic-mgmt regression to make sure there is no error


vivekrnv (approved):

paulmenzel:
Thank you for your patch. On what platform is this, and what TPM is used?

> To suppress this error log, we need set TPM module to m instead of y so that we can use kernel option to disable TPM module if need.

Why do you want to disable the TPM? As an alternative, you could use `initcall_blacklist=` with some TPM init function.

patch/kconfig-exclusions
```diff
@@ -15,6 +15,7 @@ CONFIG_MTD_SPI_NOR_USE_4K_SECTORS
 # Unset X86_PAT according to Broadcom's requirement
 CONFIG_X86_PAT
 CONFIG_MLXSW_PCI
+CONFIG_IMA
```
paulmenzel, line 18: Will that affect other systems/users?

paulmenzel (changes requested):
Please elaborate in both commit messages. “Fix comment“ is too, and I have to look at the diff, and the other commit misses the commit message body.

Junchao-Mellanox:
Hi @paulmenzel , the purpose of this PR is to suppress the TPM error log on platforms that do not need TPM. The PR did not disable TPM, it only set TPM from y to m. We see this error log on Nvidia SN5600 platform.

Junchao-Mellanox:
> Please elaborate in both commit messages. “Fix comment“ is too, and I have to look at the diff, and the other commit misses the commit message body.

I did a rebase to make these two commits into one, I hope it is clear now.

paulmenzel:
> Hi @paulmenzel , the purpose of this PR is to suppress the TPM error log on platforms that do not need TPM. The PR did not disable TPM, it only set TPM from y to m. `initcall_blacklist=` cannot work if the kernel module is a built-in(y) one.

Are you sure?

>         initcall_blacklist=  [KNL] Do not execute a comma-separated list of
>                         initcall functions.  Useful for debugging built-in
>                         modules and initcalls.

> We see this error log on Nvidia SN5600 platform.

    ERR kernel: [    3.434153] tpm tpm0: [Firmware Bug]: TPM interrupt not working, polling instead

So the device does have a TPM according to the log, but interrupts are not working. Please fix the firmware and/or Linux kernel.

patch/kconfig-exclusions
Junchao-Mellanox: actually, I am not sure about this. We need other vendor to review this change.

Junchao-Mellanox:
Hi @paulmenzel , let me clarify one thing. The PR does NOT disable TPM. It only sets TPM from `y` to `m`. But like you commented, the PR disables IMA which needs vendors to review it.

I will try your suggestion `initcall_blacklist` to disable TPM.

Junchao-Mellanox:
Hi @paulmenzel , I completely agree with you that it would be better to fix the linux kernel. But I don't think kernel maintainer is going to fix it: https://bugzilla.kernel.org/show_bug.cgi?id=204121

paulmenzel:
The Linux Kernel Bugzilla is not listed in `MAINTAINERS`, so probably ignored:

```
TPM DEVICE DRIVER
M:      Peter Huewe <peterhuewe@gmx.de>
M:      Jarkko Sakkinen <jarkko@kernel.org>
R:      Jason Gunthorpe <jgg@ziepe.ca>
L:      linux-integrity@vger.kernel.org
S:      Maintained
W:      https://codeberg.org/jarkko/linux-tpmdd-test
Q:      https://patchwork.kernel.org/project/linux-integrity/list/
T:      git git://git.kernel.org/pub/scm/linux/kernel/git/jarkko/linux-tpmdd.git
F:      Documentation/devicetree/bindings/tpm/
F:      drivers/char/tpm/
```

It’s very likely a *firmware* issue, and you have full control over your devices. Please analyze it and, if true, fix the firmware. Currently you put work on other folks to verify your changes, and it’d be great if you tried to fix your platform first. Mellanox/Nvidia should have enough engineers to look into this.

Junchao-Mellanox:
Before I raise this bug, I already consulted our kernel engineer, and I confirmed that it is not a firmware bug of Nvidia side.

paulmenzel:
> Before I raise this PR, I already consulted our kernel engineer, and I confirmed that it is not a firmware bug of Nvidia side.

Skimming the code, that can’t be true. The code checks, if the interrupts are handled, and they are not:

```
static irqreturn_t tis_int_handler(int dummy, void *dev_id)
{
        struct tpm_chip *chip = dev_id;
        struct tpm_tis_data *priv = dev_get_drvdata(&chip->dev);
        u32 interrupt;
        int rc;

        rc = tpm_tis_read32(priv, TPM_INT_STATUS(priv->locality), &interrupt);
        if (rc < 0)
                goto err;

        if (interrupt == 0)
                goto err;

        set_bit(TPM_TIS_IRQ_TESTED, &priv->flags);
        if (interrupt & TPM_INTF_DATA_AVAIL_INT)
                wake_up_interruptible(&priv->read_queue);

        if (interrupt &
            (TPM_INTF_LOCALITY_CHANGE_INT | TPM_INTF_STS_VALID_INT |
             TPM_INTF_CMD_READY_INT))
                wake_up_interruptible(&priv->int_queue);
        
        /* Clear interrupts handled with TPM_EOI */
        tpm_tis_request_locality(chip, 0);
        rc = tpm_tis_write32(priv, TPM_INT_STATUS(priv->locality), interrupt);
        tpm_tis_relinquish_locality(chip, 0);
        if (rc < 0)
                goto err;
                
        tpm_tis_read32(priv, TPM_INT_STATUS(priv->locality), &interrupt);
        return IRQ_HANDLED;

err:
        return tpm_tis_update_unhandled_irqs(chip);
}
```

Please contact the maintainers per email according to `MAINTAINERS` (Peter, Jarkko and the list linux-<integrity@vger.kernel.org>) with the engineer in Cc. I recommend to also test the latest Linux kernel (for example 6.11-rc4) to check if a possible bug in the Linux kernel was fixed in the meantime.

Junchao-Mellanox:
Hi @paulmenzel , thanks for all the information. I checked the long thread https://groups.google.com/g/fa.linux.kernel/c/aXDdaseTxOA, the maintainer Peter is also in it. And he accepted following terms (see the last reply):
```
> > When in interrupt mode, the tpm_tis module should receive an IRQ during
> > module init. Fall back to polling mode if none is received when expected.
```
So, the error log is not always a bug, and our kernel engineer has already confirmed that it is not a bug on Nvidia device. Also, I found a better way to suppress the error log when TPM is not needed: to add "tpm_tis.interrupts=0" to kernel option.

paulmenzel:
> Hi @paulmenzel , thanks for all the information. I checked the long thread https://groups.google.com/g/fa.linux.kernel/c/aXDdaseTxOA, the maintainer Peter is also in it. And he accepted following terms (see the last reply):
> 
>> When in interrupt mode, the tpm_tis module should receive an IRQ during
>> module init. Fall back to polling mode if none is received when expected.
> 
> So, the error log is not always a bug, and our kernel engineer has already confirmed that it is not a bug on Nvidia device. Also, I found a better way to suppress the error log when TPM is not needed: to add "tpm_tis.interrupts=0" to kernel option.

No. “if none is received when expected” clearly states, that is an error. Reading that thread, you *do* have a firmware bug, and the ACPI tables of the device document an interrupt for the TPM, when that interrupt does not seem to work. Please attach the output of `acpidump`, and I recommend to ask your engineers again.

Junchao-Mellanox:
sure, will confirm with him. thanks!

PR #435, opened 2024-09-23
fix driver optoe twoaddr a2h read write bug

songqh2022:
The page select register(127) only in the a0h from sfp8472 spec, so when access the a2h , the function optoe_translate_offset will update the slave_addr=a2h, and then using this slave_addr to config the page slelect register, so the contents is wrong.


songqh2022:
@paulmenzel Could you kindly help to review this request(202211 branch)?  thank you  very much!

paulmenzel:
Let’s first get https://github.com/sonic-net/sonic-linux-kernel/pull/434 reviewed and submitted.

PR #437, opened 2024-10-02
Fix RCU stall detected in kernel 6.1 when unregistering devlink

vivekrnv:
Resolves: https://github.com/sonic-net/sonic-buildimage/issues/20368

Ref: https://lore.kernel.org/stable/20241001112035.973187-1-idosch@nvidia.com/t/#u

> But first of, why? Why not just take the upstrema commits instead?
> 
> thanks,
> greg k-h

Ido:

> There were a lot of changes as part of the 6.3 cycle to completely
> rework the semantics of the devlink instance reference count. As part of
> these changes, commit d77278196441 ("devlink: bump the instance index
> directly when iterating") inadvertently fixed the bug mentioned in this
> patch. This commit cannot be applied to 6.1.y as-is because a prior
> commit (also in 6.3) moved the code to a different file (leftover.c ->
> core.c). There might be more dependencies that I'm currently unaware of.
> 
> The alternative, proposed in this patch, is to provide a minimal and
> contained fix for the bug introduced in upstream commit c2368b19807a
> ("net: devlink: introduce "unregistering" mark and use it during
> devlinks iteration") as part of the 6.0 cycle.

### DIFF
--- a/patch/0001-devlink-Fix-RCU-stall-when-unregistering-a-devlink-i.patch
+++ b/patch/0001-devlink-Fix-RCU-stall-when-unregistering-a-devlink-i.patch
@@ -1,0 +1,91 @@
+From c1bf9b90ce85dd46f958b9f0e60143d58f885963 Mon Sep 17 00:00:00 2001
+From: Ido Schimmel <idosch@nvidia.com>
+Date: Tue, 1 Oct 2024 14:20:35 +0300
+Subject: [PATCH] devlink: Fix RCU stall when unregistering a devlink instance
+
+When a devlink instance is unregistered the following happens (among
+other things):
+
+t0 - The instance is marked with 'DEVLINK_UNREGISTERING'.
+t1 - Blocking until an RCU grace period passes.
+t2 - The 'DEVLINK_UNREGISTERING' mark is cleared from the instance.
+
+When iterating over devlink instances (f.e., when requesting a dump of
+available instances) and encountering an instance that is currently
+being unregistered, the current code will loop around until the
+'DEVLINK_UNREGISTERING' mark is cleared.
+
+The iteration over devlink instances happens in an RCU critical section,
+so if the instance that is currently being unregistered was encountered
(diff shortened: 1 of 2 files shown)


dgsudharsan (approved):

saiarcot895 (approved):
Upstream kernel maintainers are in favor of backporting changes from master branch to the 6.1 stable branch instead of having a standalone/new commits on a stable branch (which makes sense in terms of maintenance and stability). However, this may mean backporting an unknown number of patches.

For the sake of simplicity, I'm in favor of backporting a single patch.

paulmenzel:
> Fix: https://github.com/sonic-net/sonic-buildimage/issues/20368

I’d use the tag *Resolves:*.

> Ref: https://lore.kernel.org/stable/20241001112035.973187-1-idosch@nvidia.com/t/#u

Please summarize the discussion. Best take @idosch’s explanation from his first reply to Greg.

vivekrnv:
> But first of, why?  Why not just take the upstrema commits instead?
> 
> thanks,
> greg k-h

Ido:
> There were a lot of changes as part of the 6.3 cycle to completely
> rework the semantics of the devlink instance reference count. As part of
> these changes, commit d77278196441 ("devlink: bump the instance index
> directly when iterating") inadvertently fixed the bug mentioned in this
> patch. This commit cannot be applied to 6.1.y as-is because a prior
> commit (also in 6.3) moved the code to a different file (leftover.c ->
> core.c). There might be more dependencies that I'm currently unaware of.
> 
> The alternative, proposed in this patch, is to provide a minimal and
> contained fix for the bug introduced in upstream commit c2368b19807a
> ("net: devlink: introduce "unregistering" mark and use it during
> devlinks iteration") as part of the 6.0 cycle.

Summary of https://lore.kernel.org/stable/20241001112035.973187-1-idosch@nvidia.com/t/#u

vivekrnv:
@bingwang-ms Can this me merged?

PR #438, opened 2024-10-03
Disable AMD memory encryption

Staphylo:
Under some specific circumstances, having memory encryption turned on lead to silent kernel crashes.
This is currently preventing SONiC using linux 6.1+ from being kexec'ed from Aboot uing kexec-tools in 64 bit.

Bisecting the kernel code pointed the following change as the culprit
[c01fce9cef8491974f7f007f90281f1608400768] x86/compressed: Add SEV-SNP feature detection/setup


saiarcot895 (approved):

Staphylo:
@saiarcot895 could this be cherry-picked to 202405 which is the first release introducing the 6.1 kernel which exposes this issue ?

PR #466, opened 2025-02-25
Fix error for writing page select register on certain xcvrs

byu343:
On Arista platforms, we see the error like below for certain type of xcvrs.
WARNING kernel: [13095.179904] scd 0000:03:00.0: #3 rsp { .reg=0x00002400, .fe=0, .foe=0, .ss=00, .ti=02, .flushed=0, .ack_error=1, .timeout_error=0, .bus_conflict_error=0, .d=0x00 } bus=41 addr=0x50 ti=2 err=-5 
ERR kernel: [13095.181197] optoe 41-0050: Non-SFP write page register for page 0 failed ret:-110!

The patch writes page select register to 0 always. However, certain type of xcvrs is not responding to it resulting in errors. The workaround here is to only write it to 0 when the existing value is not 0.


saiarcot895:
@byu343 Please see #467

PR #467, opened 2025-02-25
Revert "[optoe] Reset page select byte to 0 before upper memory access on page 0h"

prgeor:
Reverts sonic-net/sonic-linux-kernel#463


saiarcot895 (approved):
This PR is being reverted as there were issues seen with certain transceivers. This logic will be handled in userspace instead.

PR #473, opened 2025-04-02
optoe: Add CMIS Bank support for transceivers with >8 lanes

ishidawataru:
This patch adds CMIS Bank support to the 'optoe3' device class in order
to enable access to CMIS transceivers with more than 8 lanes (e.g., OSFP-XD, CPO OEs).

- Bank support is enabled only for the 'optoe3' dev class.
- ~~The default bank size is set to 4, and can be modified via the 'bank_size' sysfs entry.~~
- ~~For 'optoe3', the 'write_max' value is updated to 2 to comply with CMIS requirements,
  which mandate that both bank and page values be updated in a single WRITE operation.~~

*Updated the behavior as below after discussing with @prgeor offline*

- 'bank_size' sysfs entry is added to enable and configure the size of the bank.
- By default, bank size is set to 0.
- When enabling bank by setting a value greater than 1, the 'write_max' value is
  automatically updated to 2 to comply with CMIS requirements,
  which mandate that both bank and page values be updated in a single WRITE operation.
- If the 'write_max' value is already greater than 2, this module keeps the value as is.

Only tested with the SONiC VM + i2c-stub.

```bash
admin@sonic:~$ cat /sys/class/i2c-dev/i2c-1/device/1-0050/bank_size
4
admin@sonic:~$ hexdump /sys/class/i2c-dev/i2c-1/device/1-0050/eeprom 
0000000 0000 0000 0000 0000 0000 0000 0000 0000
*
001e880
admin@sonic:~$ echo 1 | sudo tee /sys/class/i2c-dev/i2c-1/device/1-0050/bank_size
1
admin@sonic:~$ hexdump /sys/class/i2c-dev/i2c-1/device/1-0050/eeprom 
0000000 0000 0000 0000 0000 0000 0000 0000 0000
*
0008080
```

Snip of dmesg. See bank:3 is accessed.

```
[140341.928384] optoe 1-0050: eeprom write 2@126 --> 2 (4329977748)
[140341.928386] optoe 1-0050: sff_r/w: off 122880, len 2176, chunk_start_offset 124928, chunk_offset 124928, chunk_len 128, pending_len 128
[140341.928388] optoe 1-0050: optoe_eeprom_update_client off 124928 bank:3 page:255 phy_offset:128, count:128, opcode:0
[140341.928391] optoe 1-0050: eeprom write 2@126 --> 2 (4329977748)
[140341.928393] optoe 1-0050: eeprom read 32@128 --> 32 (4329977748)
[140341.928395] optoe 1-0050: eeprom read 32@160 --> 32 (4329977748)
[140341.928397] optoe 1-0050: eeprom read 32@192 --> 32 (4329977748)
[140341.928399] optoe 1-0050: eeprom read 32@224 --> 32 (4329977748)
[140341.928401] optoe 1-0050: eeprom write 2@126 --> 2 (4329977748)
[140341.928413] optoe 1-0050: optoe_read_write: off 125056  len:4096, opcode:r
```

Not tested with real hardware yet.

### DIFF
--- a/patches-sonic/driver-support-optoe-bank-support.patch
+++ b/patches-sonic/driver-support-optoe-bank-support.patch
@@ -1,4 +1,4 @@
-From c11b89d58982c2561212f9659d9761399460263a Mon Sep 17 00:00:00 2001
+From 526a41f1fc717ddb6cfd600c899254d582735523 Mon Sep 17 00:00:00 2001
 From: Wataru Ishida <wataru.ishid@gmail.com>
 Date: Sun, 15 Feb 2026 22:34:29 +0900
 Subject: [PATCH] optoe: Add CMIS Bank support for transceivers with >8 lanes
@@ -13,11 +13,11 @@ to enable access to CMIS transceivers with more than 8 lanes (e.g., OSFP-XD, CPO
 Signed-off-by: Wataru Ishida <wataru.ishid@gmail.com>
 
 ---
- drivers/misc/eeprom/optoe.c | 183 +++++++++++++++++++++++++++++++-----
- 1 file changed, 157 insertions(+), 26 deletions(-)
+ drivers/misc/eeprom/optoe.c | 194 +++++++++++++++++++++++++++++++-----
+ 1 file changed, 167 insertions(+), 27 deletions(-)
 
 diff --git a/drivers/misc/eeprom/optoe.c b/drivers/misc/eeprom/optoe.c
-index ba4ca17..97bde31 100644
+index ba4ca17..8ba71ef 100644
 --- a/drivers/misc/eeprom/optoe.c
 +++ b/drivers/misc/eeprom/optoe.c
 @@ -101,9 +101,50 @@
@@ -98,15 +98,21 @@ index ba4ca17..97bde31 100644
  #define OPTOE_PAGE_SELECT_REG   0x7F
  #define ONE_ADDR_PAGEABLE_REG 0x02
  #define QSFP_NOT_PAGEABLE (1<<2)
-@@ -195,6 +247,7 @@ struct optoe_data {
+@@ -195,6 +247,13 @@ struct optoe_data {
  	u8 *writebuf;
  	unsigned int write_max;
  	unsigned int write_timeout;
 +	unsigned int bank_size; /* 0 means bank is not supported */
++
++	/* Indicates if page restore has failed.
++	 * If true, the driver doesn't skip writing to page select register
++	 * even for acesses to page 0.
++	 */
++	bool page_restore_failed;
  
  	unsigned int num_addresses;
  
-@@ -245,13 +298,19 @@ static const struct i2c_device_id optoe_ids[] = {
+@@ -245,13 +304,19 @@ static const struct i2c_device_id optoe_ids[] = {
  };
  MODULE_DEVICE_TABLE(i2c, optoe_ids);
  
@@ -127,7 +133,7 @@ index ba4ca17..97bde31 100644
   *
   * Handles both single address (eg QSFP) and two address (eg SFP).
   *     For SFP, offset 0-255 are on client[0], >255 is on client[1]
-@@ -274,7 +333,7 @@ MODULE_DEVICE_TABLE(i2c, optoe_ids);
+@@ -274,7 +339,7 @@ MODULE_DEVICE_TABLE(i2c, optoe_ids);
   */
  
  static uint8_t optoe_translate_offset(struct optoe_data *optoe,
@@ -136,7 +142,7 @@ index ba4ca17..97bde31 100644
  {
  	unsigned int page = 0;
  
-@@ -297,12 +356,17 @@ static uint8_t optoe_translate_offset(struct optoe_data *optoe,
+@@ -297,12 +362,17 @@ static uint8_t optoe_translate_offset(struct optoe_data *optoe,
  	if (*offset < OPTOE_PAGE_SIZE)
  		return page;
  
@@ -159,7 +165,7 @@ index ba4ca17..97bde31 100644
  }
  
  static ssize_t optoe_eeprom_read(struct optoe_data *optoe,
-@@ -511,21 +575,38 @@ static ssize_t optoe_eeprom_update_client(struct optoe_data *optoe,
+@@ -511,23 +581,41 @@ static ssize_t optoe_eeprom_update_client(struct optoe_data *optoe,
  {
  	struct i2c_client *client;
  	ssize_t retval = 0;
@@ -173,11 +179,12 @@ index ba4ca17..97bde31 100644
  	dev_dbg(&client->dev,
 -		"%s off %lld  page:%d phy_offset:%lld, count:%ld, opcode:%d\n",
 -		__func__, off, page, phy_offset, (long int) count, opcode);
-+		"%s off %lld bank:%d page:%d phy_offset:%lld, count:%ld, opcode:%d\n",
-+		__func__, off, bank, page, phy_offset, (long int) count, opcode);
- 	if (page > 0) {
+-	if (page > 0) {
 -		ret = optoe_eeprom_write(optoe, client, &page,
 -			OPTOE_PAGE_SELECT_REG, 1);
++		"%s off %lld bank:%d page:%d phy_offset:%lld, count:%ld, opcode:%d\n",
++		__func__, off, bank, page, phy_offset, (long int) count, opcode);
++	if (optoe->page_restore_failed || page > 0) {
 +		/*
 +		 * CMIS 5.3 8.2.15 Page Mapping
 +		 *
@@ -205,8 +212,11 @@ index ba4ca17..97bde31 100644
 +					bank, page, ret);
  			return ret;
  		}
++		optoe->page_restore_failed = false;
  	}
-@@ -553,13 +634,18 @@ static ssize_t optoe_eeprom_update_client(struct optoe_data *optoe,
+ 
+ 	while (count) {
+@@ -553,13 +641,20 @@ static ssize_t optoe_eeprom_update_client(struct optoe_data *optoe,
  
  
  	if (page > 0) {
@@ -227,10 +237,12 @@ index ba4ca17..97bde31 100644
  			dev_err(&client->dev,
 -				"Restore page register to 0 failed:%d!\n", ret);
 +				"Restore bank, page register to (0, 0) failed:%d!\n", ret);
++			optoe->page_restore_failed = true;
++
  			/* error only if nothing has been transferred */
  			if (retval == 0)
  				retval = ret;
-@@ -622,8 +708,10 @@ static ssize_t optoe_page_legal(struct optoe_data *optoe,
+@@ -622,8 +717,10 @@ static ssize_t optoe_page_legal(struct optoe_data *optoe,
  		/* if no pages needed, we're good */
  		if ((off + len) <= ONE_ADDR_EEPROM_UNPAGED_SIZE)
  			return len;
@@ -242,7 +254,7 @@ index ba4ca17..97bde31 100644
  			return OPTOE_EOF;
  		/* in between, are pages supported? */
  		status = optoe_eeprom_read(optoe, client, &regval,
-@@ -665,7 +753,7 @@ static ssize_t optoe_page_legal(struct optoe_data *optoe,
+@@ -665,7 +762,7 @@ static ssize_t optoe_page_legal(struct optoe_data *optoe,
  			maxlen = ONE_ADDR_EEPROM_UNPAGED_SIZE - off;
  		} else {
  			/* Pages supported, trim len to the end of pages */
@@ -251,7 +263,7 @@ index ba4ca17..97bde31 100644
  		}
  		len = (len > maxlen) ? maxlen : len;
  		dev_dbg(&client->dev,
-@@ -995,6 +1083,47 @@ static ssize_t set_port_name(struct device *dev,
+@@ -995,6 +1092,47 @@ static ssize_t set_port_name(struct device *dev,
  	return count;
  }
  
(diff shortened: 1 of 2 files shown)


ishidawataru:
@prgeor PTAL

ishidawataru:
The optoe linear address space is extended as below to support Bank without changing how the user space interact with optoe.

Screenshot 2025-04-03 at 9 58 47

patch/driver-support-optoe-bank-support.patch
```diff
@@ -0,0 +1,307 @@
...
+ 	int num_addresses = 0;
+ 	char port_name[MAX_PORT_NAME_LEN];
++	/*
++	 * NOTE: AN-2079
++	 * Finisar recommends that the host implement 1 byte writes
++	 * only since this module only supports 32 byte page boundaries.
++	 * 2 byte writes are acceptable for PE and Vout changes per
++	 * Application Note AN-2071.
++	 *
++	 * use 2 bytes for CMIS devices since CMIS 5.3 requires write both BankSelect and PageSlect in one WRITE acess.
```
paulmenzel, line 255: Wrap the line?

patch/driver-support-optoe-bank-support.patch
```diff
@@ -0,0 +1,307 @@
...
+Subject: [PATCH] optoe: Add CMIS Bank support for transceivers with >8 lanes
+
+This patch adds CMIS Bank support to the 'optoe3' device class in order
+to enable access to CMIS transceivers with more than 8 lanes (e.g., OSFP-XD, CPO OEs).
+
+- Bank support is enabled only for the 'optoe3' dev class.
+- The default bank size is set to 4, and can be modified via the 'bank_size' sysfs entry.
+- For 'optoe3', the 'write_max' value is updated to 2 to comply with CMIS requirements,
+  which mandate that both bank and page values be updated in a single WRITE operation.
+
```
paulmenzel, line 13: It’d be great if you could add a paragraph how to verify/test this to the commit message too.

paulmenzel (commented):
Looks great overall.

ishidawataru:
@prgeor I updated the patch based on our discussion PTAL.

prgeor:
> @prgeor I updated the patch based on our discussion PTAL.

@ishidawataru Ack. reviewing.

prgeor:
@ishidawataru can you define what do you mean by `bank_size` ?

ishidawataru:
@prgeor 

`bank_size` represents the number of CMIS transceiver banks that the optoe driver assumes.
This setting is necessary because `optoe` maps CMIS 3D addresses into a linear address space.
The maximum of the linear address space is determined based on the `bank_size`, and address validity is checked accordingly.

For example, if `bank_size` is 0 or 1, the linear address space ends at 0x8080.
If `bank_size` is 2, the address space extends up to 0xf880.

```
admin@sonic:~$ echo 1 | sudo tee /sys/class/i2c-dev/i2c-1/device/1-0050/bank_size                                                                                      
1
admin@sonic:~$ hexdump /sys/class/i2c-dev/i2c-1/device/1-0050/eeprom                                                                                       
0000000 0000 0000 0000 0000 0000 0000 0000 0000
*
0008080
admin@sonic:~$ echo 2 | sudo tee /sys/class/i2c-dev/i2c-1/device/1-0050/bank_size                                                                                      
2
admin@sonic:~$ hexdump /sys/class/i2c-dev/i2c-1/device/1-0050/eeprom                                                                                       
0000000 0000 0000 0000 0000 0000 0000 0000 0000
*
000f880
```

I also added a commit that explains this in the comment.

https://github.com/sonic-net/sonic-linux-kernel/pull/473/commits/1b2fb60f6e9e0d993ea8cafe01017f9dfdfcc48b

patch/driver-support-optoe-bank-support.patch
```diff
@@ -0,0 +1,329 @@
...
++		loff_t *offset, struct i2c_client **client, uint8_t *bank)
+ {
+ 	unsigned int page = 0;
+ 
+@@ -301,8 +352,9 @@ static uint8_t optoe_translate_offset(struct optoe_data *optoe,
+ 	page = (*offset >> 7)-1;
+ 	/* 0x80 places the offset in the top half, offset is last 7 bits */
+ 	*offset = OPTOE_PAGE_SIZE + (*offset & 0x7f);
+-
+-	return page;  /* note also returning client and offset */
```
prgeor, line 139: @ishidawataru The computation does not consider full range of linear address for all banks. Its OK to use all page address for bank 0 i.e Assume that page 0x10, 0x11, 0x12... 0xFF exist for Bank 0 as well. 

So that the [getaddr()](https://github.com/sonic-net/sonic-platform-common/blob/dd6cc6487b32f1d9b7a5f9925b0d599223bc7f4a/sonic_platform_base/sonic_xcvr/mem_maps/public/cmis.py#L115) can be implemented as follows:-
```
    def getaddr(self, bank=0, page, offset, page_size=128):
          return ((uint32_t)bank * PAGES_PER_BANK + page) * page_size + offset;
```

In your case you are not wasting the address space for `OPTOE_NON_BANKED_PAGE_SIZE`  but its OK to loose those many address range to keep the use space implementation of `getaddr()` simple

Here is the simple decoding of the above linear address that optoe needs to do:-

```
static uint8_t optoe_translate_offset(struct optoe_data *optoe,
		loff_t *offset, struct i2c_client **client, uint8_t *bank)
{
           unsigned int page = 0;
           
          *bank = *offset / (PAGES_PER_BANK * PAGE_SIZE);
           page = ((*offset / PAGE_SIZE) % PAGES_PER_BANK) - 1;
          *offset = PAGE_SIZE + linear_address % PAGE_SIZE;

         return page;
}
```

Assumption here is the user application will provide only the valid pages and the module will return IO error if incorrect Page address is selected.

patch/driver-support-optoe-bank-support.patch
```diff
@@ -0,0 +1,329 @@
...
+@@ -101,9 +101,42 @@
+  * considerably more pages (at least to page 0xAF), which this driver
+  * supports.
+  *
+- * NOTE: This version of the driver ONLY SUPPORTS BANK 0 PAGES on CMIS
+- * devices.
++ * For CMIS transceivers that support Banked Pages, access to these pages
++ * is also supported. To access the banked pages, set the number of banks
++ * to access via the `bank_size` sysfs entry.
++ * By default, `bank_size` is set to 0, which disables this feature.
```
prgeor, line 35: @ishidawataru We probably don't need this. see my comments below.

patch/driver-support-optoe-bank-support.patch
```diff
@@ -0,0 +1,280 @@
...
+diff --git a/drivers/misc/eeprom/optoe.c b/drivers/misc/eeprom/optoe.c
+index 22d2c0cd4..71bd90204 100644
+--- a/drivers/misc/eeprom/optoe.c
++++ b/drivers/misc/eeprom/optoe.c
+@@ -150,6 +150,12 @@ struct optoe_platform_data {
+ 
+ /* fundamental unit of addressing for EEPROM */
+ #define OPTOE_PAGE_SIZE 128
++
++#define OPTOE_DEFAULT_BANK_SIZE       0
```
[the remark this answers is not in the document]
prgeor: @ishidawataru This probably can be removed if we implement full range of linear address for all bank pages. See my comments below

patch/driver-support-optoe-bank-support.patch
ishidawataru: @prgeor  [This](https://github.com/sonic-net/sonic-linux-kernel/pull/473#issuecomment-2774071802) is the linear address map of the current implementation.

Are you suggesting to change it like below?

Screenshot 2025-04-24 at 18 04 36

Returning I/O errors for the grayed-out area will prevent us from using cat on the EEPROM file as shown below.

```
admin@sonic:~$ hexdump /sys/class/i2c-dev/i2c-1/device/1-0050/eeprom 
0000000 0000 0000 0000 0000 0000 0000 0000 0000
*
001e880
```

I suggest keeping the current implementation or returning the value from bank 0, rather than returning I/O errors.

What do you think?

patch/driver-support-optoe-bank-support.patch
```diff
@@ -0,0 +1,280 @@
...
++#define OPTOE_MAX_SUPPORTED_BANK_SIZE 8
++#define OPTOE_NON_BANKED_PAGE_SIZE    16  /* page 00h-0Fh are not banked */
++#define OPTOE_BANKED_PAGE_SIZE        240 /* page 10h-FFh are banked */
```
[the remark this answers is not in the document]
ishidawataru: If these constants are still necessary after the discussion below, I’ll rename them as you suggested.

patch/driver-support-optoe-bank-support.patch
prgeor: @ishidawataru I am not sure if `hexdump` will be user friendly to read so many pages across multiple banks for the user to scroll and look for. Alternatively, one can use the userspace tools like `sfputil` to dump only relavant page from a particular bank.

patch/driver-support-optoe-bank-support.patch
```diff
@@ -0,0 +1,329 @@
...
++	ssize_t count;
++
++	mutex_lock(&optoe->lock);
++	count = sprintf(buf, "%u\n", optoe->bank_size);
++	mutex_unlock(&optoe->lock);
++
++	return count;
++}
++
++static ssize_t set_bank_size(struct device *dev,
```
prgeor: @ishidawataru do we need this? What is the usecase? I thought the number of banks is already advertised so the optoe driver can read the advertisement and cache the value? Are you considering a case where the module is swapped?

patches-sonic/driver-support-optoe-bank-support.patch
```diff
@@ -0,0 +1,360 @@
...
++		 * For just a Page Index change (mapping another Page in the current Bank),
++		 * or for mapping an arbitrary unbanked Page to Upper Memory, a host may WRITE only the PageSelect Byte.
++		 */
++		if (bank > 0) {
++			char buf[2] = {bank, page};
++			unsigned int write_max_orig = optoe->write_max;
++			optoe->write_max = 2; /* ensure both bank and page are written together */
++			ret = optoe_eeprom_write(optoe, client, buf,
++				OPTOE_BANK_SELECT_REG, 2);
++			optoe->write_max = write_max_orig;
```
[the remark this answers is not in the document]
ishidawataru: `optoe_eeprom_update_client()` is called inside `optoe_read_write()`, where the lock is already taken.

patches-sonic/driver-support-optoe-bank-support.patch
```diff
@@ -0,0 +1,360 @@
...
++	struct optoe_data *optoe = i2c_get_clientdata(client);
++	unsigned int bank_size;
++
++	// setting bank size is only supported for the CMIS device
++	if (optoe->dev_class != CMIS_ADDR) {
++		return -EINVAL;
++	}
++
++	if (kstrtouint(buf, 0, &bank_size) != 0 ||
++		bank_size < 0 || bank_size > OPTOE_MAX_SUPPORTED_BANK_SIZE)
```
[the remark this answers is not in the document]
ishidawataru: addressed by https://github.com/sonic-net/sonic-linux-kernel/pull/473/changes/424ec6c9e06b33766700c2fdb3627f47bbff85be

mihirpat1 (dismissed):

bobby-nexthop (approved):
Thank you!

patches-sonic/driver-support-optoe-bank-support.patch
```diff
@@ -0,0 +1,361 @@
...
+ 	&dev_attr_dev_class.attr,
++	&dev_attr_bank_size.attr,
+ 	NULL,
+ };
+ 
+@@ -1027,7 +1171,6 @@ static int optoe_probe(struct i2c_client *client)
+ 	const struct i2c_device_id *id;
+ 	int num_addresses = 0;
+ 	char port_name[MAX_PORT_NAME_LEN];
+-
```
prgeor: @ishidawataru can you revert this?

ishidawataru: addressed in this commit: https://github.com/sonic-net/sonic-linux-kernel/pull/473/changes/a2d8ba6b7d5b9b506a804caedb9850865abf1f69

ishidawataru:
@prgeor all CI checks have passed. PTAL

prgeor (dismissed):

patches-sonic/driver-support-optoe-bank-support.patch
```diff
@@ -0,0 +1,353 @@
...
++
++/*
++ * By default, banked pages are not supported (bank_size = 0), which limits
++ * the maximum linear address space to 256 pages (32KB). When bank_size is
++ * greater than 0, the driver supports access to additional banked pages, with
++ * the total address space increasing by 256 pages (32KB) for each additional bank.
++ * The maximum supported bank size in this version is 8.
++ */
++#define OPTOE_DEFAULT_BANK_SIZE       0
++#define OPTOE_MAX_SUPPORTED_BANK_SIZE 8
```
prgeor: @ishidawataru Why the value is 8? As per CMIS spec, the max number of lanes is 32 which means MAX bank size is 4

Image

aditya-nexthop: We should probably update to 16 to allow NAD

image

micas-net:
I am doing local integrating, once the functionality be verified, will merge this! @prgeor

patches-sonic/driver-support-optoe-bank-support.patch
```diff
@@ -0,0 +1,329 @@
...
++		 *
++		 * For an arbitrary Page Address change or for just a Bank Index change,
++		 * a host must write both BankSelect and PageSelect in one WRITE access,
++		 * even if the PageSelect value does not change.
++		 * The module does not begin processing the BankSelect value until after the PageSelect register has been written.
++		 *
++		 * For just a Page Index change (mapping another Page in the current Bank),
++		 * or for mapping an arbitrary unbanked Page to Upper Memory, a host may WRITE only the PageSelect Byte.
++		 */
++		if (bank > 0) {
```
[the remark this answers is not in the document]
aditya-nexthop: Maybe we shouldn't incorporate this check because bank size can be determined not just by this register but also by NAD support. Maybe best to leave check to userspace?

aditya-nexthop (approved):
Thank you for patiently working on these changes @ishidawataru

prgeor (approved):

PR #478, opened 2025-04-29
Enable CONFIG_RTC_INTF_DEV_UIE_EMUL to enable clock interrupt emulation

saiarcot895:
SONiC has switched to using Chrony as the NTP daemon and to manage the hardware clock (RTC). One thing that Chrony requires for the purposes of drift tracking is the ability for the RTC to generate an interrupt whenever it changes. For most hardware that exposes a RTC chip to the kernel, this is fine, as this is well supported. However, for hardware that instead expose the hardware clock only via EFI, these update interrupts cannot be generated.

As a workaround, for RTC chips/drivers that don't support update interrupts, the kernel supports reading the time from the hardware clock several times per second and then faking an interrupt based on what it reads. Based on a read of the kernel code, this codepath is taken only if the RTC driver doesn't claim to support update interrupts or alarms, so for most hardware, this should have no impact on performance. For hardware that only exposes the hardware clock via EFI, there might be a slight impact; however, such devices should be few in number.


vvolam (approved):
LGTM, Thank you

PR #481, opened 2025-05-19
[Nexthop] Add PCI_STUB and ADM1266 to kconfig-inclusions

michaelc-nexthop:
Add Linux kernel patches for Nexthop Platform Support

Enable the following kernel configuration options:

- CONFIG_PCI_STUB=y
- CONFIG_SENSORS_ADM1266=m

Rationale:

- PCI_STUB is required to allow manual control over PCI device binding. In Nexthop’s system, this is used to bind an unused AMD XGBE MAC to the pci-stub driver, preventing the amd-xgbe driver from claiming it.

- ADM1266 is needed to support the ADM1266 PMBus device now present on Nexthop hardware. This driver is used for monitoring voltage sensors on the system.


patch/0001-Add-support-for-new-qspi-macronix-flash-mx77u25650f.patch
```diff
@@ -0,0 +1,29 @@
+From d00e6b6d68f39fe09a9d0062681ae1a666e03fcc Mon Sep 17 00:00:00 2001
+From: Nate White <nate@nexthop.ai>
+Date: Thu, 13 Mar 2025 13:42:08 +0000
+Subject: [PATCH] Add support for new qspi macronix flash mx77u25650f
```
paulmenzel, line 4: Has this been sent upstream yet?

I found [\[PATCH\] mtd: spi-nor: macronix: add MX77U25650F, MX77U25655F and MX77U51250F](https://lore.kernel.org/all/20250317182058.2060277-1-czapiga@google.com/#r).

paulmenzel (commented):
Three commits would be nice with a good commit message. Why do you need `PCI_STUB`, for example?

patch/0001-Add-support-for-new-qspi-macronix-flash-mx77u25650f.patch
nate-nexthop: We didn't spend the time upstreaming this patch for a couple of reasons. As you point out there is at least one patch already sent to upstream this part. Also, the spi-nor subsystem has gone through a refactor after v6.1, which changes a few things including the macros used to define these part lists. There is also now an entry that just matches the Macronix vendor ID, so it's possible this part will just work with its SFDP in future versions.

All that being said, this patch isn't critical for our device and we aren't ready to upstream the code that needs it, so we've withdrawn this part of the PR.

Thanks for reviewing!

lotus-nexthop (approved):

saiarcot895 (approved):

PR #484, opened 2025-05-28
Add SHA512 algorithm

gpunathilell:
Add CONFIG_CRYPTO_SHA512  kernel configuration to the default kconfig inclusions, This is used to prevent the following error logs:
```
ERR kernel: [    0.922382] ima: Can not allocate sha384 (reason: -2)
ERR kernel: [    1.047402] ima: No suitable TPM algorithm for boot aggregate
```
On systems which use the SHA384/512 algorithms


paulmenzel (commented):
sha → SHA in the summary/title?

The rest looks good, though I’d add a commit message to the commit, but GitHub is adding the merge/pull request summary with the current settings (that squashes all commits).

PR #488, opened 2025-07-02
Update to 6.12.30 for Debian Trixie

saiarcot895:


paulmenzel (commented):
Nice. Thank you for doing this. It’d be great if you elaborated more about the new build infrastructure, or even made that a separate merge/pull request if possible.

liushilongbuaa:
failed on disk full. Let me update agent pool.

saiarcot895:
I didn't include the symlink to the data directory in the agent pool change. I think that would fix it?

liushilongbuaa:
> I didn't include the symlink to the data directory in the agent pool change. I think that would fix it?

that would fix. Maybe it's better to change agent instead of pipeline.

PR #509, opened 2025-08-12
Add upstream fix for bridge link show failure

tirupatihemanth:
Backport https://github.com/torvalds/linux/commit/e92df790d07a8eea873efcb84776e7b71f81c7d5 into sonic-linux-kernel
On SmartSwitch we see the following error when hostside PF interfaces (for DPUs) created by mlx5_core driver are present.

```
root@sonic:/home/admin/# bridge vlan show
port              vlan-id
***                ***
RTNETLINK answers: Operation not permitted
Dump terminated
```
Similar error is seen for bridge link show command as well.
This backport is expected to fix these issues.

### DIFF
--- a/patch/series
+++ b/patch/series
@@ -71,6 +71,8 @@ Support-for-fullcone-nat.patch
 #
 
 0001-efi-memattr-Ignore-table-if-the-size-is-clearly-bogus.patch
+# Backport from v6.14 to v6.1.132 and v6.12.20
+0001-prevent-bridge-link-show-failure-for-non-eswitch-allowed-devices.patch
 
 # Mellanox patches for 5.10
 ###-> mellanox_sdk-start
--- a/patch/0001-prevent-bridge-link-show-failure-for-non-eswitch-allowed-devices.patch
+++ b/patch/0001-prevent-bridge-link-show-failure-for-non-eswitch-allowed-devices.patch
@@ -1,0 +1,45 @@
+From e92df790d07a8eea873efcb84776e7b71f81c7d5 Mon Sep 17 00:00:00 2001
+From: Carolina Jubran <cjubran@nvidia.com>
+Date: Tue, 11 Mar 2025 00:01:44 +0200
+Subject: [PATCH] net/mlx5e: Prevent bridge link show failure for
+ non-eswitch-allowed devices
+
+mlx5_eswitch_get_vepa returns -EPERM if the device lacks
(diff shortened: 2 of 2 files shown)


patch/series
```diff
@@ -70,6 +70,7 @@ Support-for-fullcone-nat.patch
 # Tbridge-per-port-multicast-broadcast-flood-flags.patch
 #
 
+0001-prevent-bridge-link-show-failure.patch
```
paulmenzel, line 73: Also, please use the name created by `git format-patch`, so that it contains the whole subject line.

tirupatihemanth: Updated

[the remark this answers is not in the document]
tirupatihemanth: Added related backport related information

patch/series
```diff
@@ -71,6 +71,8 @@ Support-for-fullcone-nat.patch
 #
 
 0001-efi-memattr-Ignore-table-if-the-size-is-clearly-bogus.patch
+# Backport from v6.14 for v6.1 and v6.12
```
paulmenzel, line 74: Thank you for adding this comment, but the patch is actually in released versions, but the SONiC one is older then that. I’d prefer to have in the comment, the releases the patch was added to the series.

paulmenzel:
(A minor comment: In my inbox your name is all upper case: HEMANTH KUMAR TIRUPATI. If possible, it’d be great, if capitalized only the first letter of each name.)

dgsudharsan (approved):

patch/series
saiarcot895: Do you mean specifying that the patch is in 6.1.132 and 6.12.20? (basically specifying the specific patch version)

Web link to make it a bit easier to find out which tags a commit has been backported to: https://kernel.dance/#e92df790d07a8eea873efcb84776e7b71f81c7d5

tirupatihemanth: Thank you @saiarcot895
I added the comment to highlight the major kernel versions of the most recent released and upcoming sonic images based on debian 12 and 13.

saiarcot895:
Required pipeline run has fully succeeded, https://dev.azure.com/mssonic/build/_build/results?buildId=929686&view=results and https://dev.azure.com/mssonic/build/_build/results?buildId=925239&view=results. It appears that the required status checker has become duplicated and/or split from the actual checker.

qiluo-msft:
Will force merge

image

qiluo-msft (approved):

PR #514, opened 2025-10-29
Add a patch for printing the AMD Zen CPU reset reason

lotus-nexthop:
Upstream commits:
- https://github.com/torvalds/linux/commit/ab8131028710d009ab93d6bffd2a2749ade909b0
- https://github.com/torvalds/linux/commit/e9576e078220c50ace9e9087355423de23e25fa5

The patch had to be adapted to v6.1 we're using, that was basically adding the entire contents (5 constants) of `fch.h` as the file didn't exist in v6.1, and updating the patch for `amd.c` for context.

#### Testing

If we intentionally trigger a CPU soft reset (with `sudo reboot -f`) I see this:
```
admin@gold208-dut:~$ sudo dmesg | grep -i reason
[    0.635233] x86/amd: Previous system reset reason [0x00080800]: software wrote 0x6 to reset control register 0xCF9
```

If we intentionally trigger the CPU FCH Watchdog, I see this:
```
admin@gold208-dut:~$ sudo dmesg | grep reason
[    0.632563] x86/amd: Previous system reset reason [0x02000800]: hardware watchdog timer expired
```

To enable watchdog we create a 
`/etc/systemd/system.conf.d/override.conf‎`
with the contents:
```
[Manager]
RuntimeWatchdogSec=default
WatchdogDevice=/dev/watchdog1
```
To trigger the watchdog:
`sudo tee /dev/watchdog1` and enter just one character and let the device be for a minute or so.


nate-nexthop:
@saiarcot895 could you take a look please?

saiarcot895 (approved):

bingwang-ms:
@lotus-nexthop Is this PR also required for 202512?

lotus-nexthop:
> @lotus-nexthop Is this PR also required for 202512?

Thanks @bingwang-ms for cherry picking this to 202512. 
Yes this is useful to have in 202512 for debugging reboot cause AMD Zen based platforms, which includes NH-4210.

PR #552, opened 2026-03-30
KEMPLD_DISABLE_ALL disable MFD-children autoinstall

yanmarkman:
# Why I did it:
The KEMPLD modprobe and I2c-KEMPLD bus to be eliminated for marvell-prestera.
But .ko modules to be present on file-system for manual install operations.

# How I did it
Tried but Not possible:
1). Classic /etc/modprobe.d/blacklist.conf is not present on pure SONiC rootfs
 Adding to platform's overlay -- present on rootfs but doesn't present on initrd.img -> not work
2). "update-initramfs -u" on already running SONiC is too long and risky
 especially for CRT security if board starts with "initial old" date-time.

So, the only option to blacklist the KEMPLD is modify kempld-core driver by eliminating
child gpio, i2c child installation.

For this request the patch does:

1). Add new flag CONFIG_KEMPLD_DISABLE_ALL.

Set =y means - no KEMPLD child devices will be auto-instantiated.
Modules will still be built (CONFIG_*_KEMPLD=m) and present on filesystem,
but nothing will be auto-loaded.

2). Set =y for marvell-prestera.

3). Expand defines.toml to handle config platform-specific (just like it is already done for armhf/arm64).


yanmarkman:
@saiarcot895 
1). files/build_templates/sonic_debian_extensions.j2
Would maintainers accept the changes in template ?!
Which .j2 variant have you meant:

1.a) create text .conf directly
if [[ $CONFIGURED_PLATFORM == marvell-prestera ]]; then
  if [[ $CONFIGURED_ARCH == amd64 ]]; then
            echo echo echo >> modprobe.d/marvell-blacklist.conf

1.b) Copy from below PLATFORM's path (seems) would fail on Azure-checker complaining about change on Read-Only
 platform/marvell-prestera/mrvl-prestera/platform/amd64/common/etc/modprobe.d/marvell-blacklist.conf



2). why should the kernel modules not get loaded? What happens if it does get loaded?
2.a) Occasional i2c-bus 1/2 enumeration swap on Kernel-startup between SMBUS and Kontron-kempld
  leading to permanent PMON failure
2.b) Marvell-amd64-Falcon device has I2C-mux collision between busses leading to stuck on bus

yanmarkman:
PR https://github.com/sonic-net/sonic-buildimage/pull/26484 changing files/build_templates/sonic_debian_extensions.j2
 is under test now. In case of success it would replace this Kernel's PR.

yanmarkman:
Closed, replaced by the
https://github.com/sonic-net/sonic-buildimage/pull/26484

PR #566, opened 2026-05-04
Integrate HW-MGMT Version 7.0060.1002

DannyIsa:
#### Why I did it

Integrate Mellanox/NVIDIA HW-MGMT version 7.0060.1002 into sonic-linux-kernel.

Source tag:
https://github.com/Mellanox/hw-mgmt/tree/V.7.0060.1002

Patch status table:
https://github.com/Mellanox/hw-mgmt/blob/V.7.0060.1002/recipes-kernel/linux/Patch_Status_Table.txt

This update refreshes the SONiC kernel patch stack and enables the required HW-MGMT amd64 modules in a separate commit.

Signed-off-by: Daniel Isakov <disakov@nvidia.com>

#### How I did it
Run make integrate-mlnx-hw-mgmt


config.local/amd64/config.sonic
```diff
@@ -83,6 +83,9 @@ CONFIG_MLXSW_CORE_HWMON=y
 CONFIG_MLXSW_I2C=m
 CONFIG_MLXSW_MINIMAL=m
 CONFIG_MLXSW_CORE_THERMAL=n
+CONFIG_NVSW_HOST_L1=m
```
paulmenzel, line 86: Please make changes in this file a separate commit with an elaborate commit message.

patches-sonic/series
```diff
@@ -111,11 +111,14 @@ driver-i2c-smbus-add-disable_spd-module-parameter.patch
...
 0054-i2c-asf-Introduce-MCTP-support-over-ASF-controller.patch
+0054-1-platform-mellanox-Downstream-Add-support-for-new-Nvi.patch
 0055-platform-mellanox-Downstream-Add-support-DGX-flavor-.patch
+0055-1-platform-mellanox-mlx-platform-Add-support-DGX-flavo.patch
 0056-hwmon-pmbus-mp2845-Add-support-for-MP2845-device.patch
 0057-hwmon-pmbus-mp5926-Add-support-for-MP5926-device.patch
+0058-PCI-DOE-Poll-DOE-Busy-bit-for-up-to-1-second-in-pci.patch
+0059-hwmon-pmbus-xdpe1a2g7b-Add-support-for-XDPE1A2G7B-5B.patch
 8000-mlxsw-Use-weak-reverse-dependencies-for-firmware-fla.patch
-8001-platform-mlx-platform-Downstream-Add-SPI-path-for-ra.patch
```
paulmenzel, line 119: The commit message should mention the renaming and the reason.

paulmenzel (changes requested):
A bot should not sign off commits.

abelamit (approved):

vivekrnv (approved):

patches-sonic/series
tirupatihemanth: Hi @paulmenzel  These changes are made when we integrate a new lower-layer component. The reason for changes can be understood in more detail from https://github.com/Mellanox/hw-mgmt repo from the specific version of the component - in this case 7.0060.1002 (mentioned in the title). As there are several changes for each component upgrade and each change has a reason from the hw-mgmt repo. As we do this upgrade often and these integration commits could get very verbose. So I believe we should keep this to minimal and abstract away the details of lower layer components - and just point to the hw-mgmt repo for all information. Let us know if we can't resolve this

config.local/amd64/config.sonic
tirupatihemanth: same reasoning as below about the commit message. Is this a nit or any specific reason behind separate commit?

patches-sonic/series
paulmenzel: @tirupatihemanth, thank you for repsonding. Reviewers time is scarce, and, therefore, review should be made as simple as possible. I went to https://github.com/Mellanox/hw-mgmt, and can’t find that tag. Also not on the first release page. One improvement would be to mention the previous version in the commit message, and also add URLs for the added release.

Your commit message already lists patches, and has some superfluous colons, which should be removed. I think that list could be expanded a little to help the reviewer wit renames and so on.

config.local/amd64/config.sonic
paulmenzel: It’s not mentioned in the commit message as far as I can see, and looks like a separate change, that should be a separate commit (or even merge/pull request) with an explanation.

DannyIsa:
@paulmenzel
Thanks, understood. I’ll respin the PR to address your comments.

saiarcot895 (approved):

DannyIsa:
@saiarcot895 
Hi thanks for the approval
Can you help me with merging the PR?

PR #569, opened 2026-05-07
Remove CONFIG_ASPEED_SSP, its specific to 2600 and not supported on 2700

chander-nexthop:
Remove this option from the aspeed kernel build


chinmoy-nexthop (approved):
LGTM

saiarcot895 (approved):

PR #576, opened 2026-05-13
[multi-asic] add drop monitor patches to support multi-asic

Yakiv-Huryk:
**What I did**
Added three downstream patches that extend Linux drop_monitor (net/core/drop_monitor.c):
* 0003-drop_monitor-Allow-running-multiple-instances-of-dropwatch.patch — reference-count software and hardware drop tracing so more than one dropwatch client can enable tracing without tripping “already enabled” errors; monitoring state is derived from those refcounts.
* 0004-drop_monitor-Send-hardware-drop-notifications-to-the.patch — multicast hardware drop notifications to the correct network namespace.
* 0005-drop_monitor-Do-not-block-a-configuration-change-if-it-is-a.patch — consolidate config handling so a no-op config (unchanged alert mode, trunc length, queue length) succeeds even while monitoring is active; real changes still return -EBUSY while monitoring.

**Why I did it**
On multi-ASIC SONIC, multiple drop monitor subscribers need to configure/subscribe without fighting the global “trace already on” / “HW monitoring already enabled” state.
Hardware drops are tied to a device’s netns; sending HW notifications only in the initial netns is wrong for per-ASIC namespaces.
Allowing idempotent config while monitoring avoids unnecessary -EBUSY when different applications resend the same parameters.

**How I verified it**
Manual functional tests in multi-ASIC SONiC (multiple drop monitor subscribers in per-ASIC namespaces) 

**Details if related**

**Patch upstreaming plan**
These patches will not be upstreamed.

### DIFF
--- a/patches-sonic/0003-drop_monitor-Allow-running-multiple-instances-of-dropwatch.patch
+++ b/patches-sonic/0003-drop_monitor-Allow-running-multiple-instances-of-dropwatch.patch
@@ -1,0 +1,146 @@
+From b1fe183ad73ab30c6ed6aef7b961e3b440dfa174 Mon Sep 17 00:00:00 2001
+From: Ido Schimmel <idosch@nvidia.com>
+Date: Wed, 11 Feb 2026 18:55:30 +0200
+Subject: [PATCH] drop_monitor: Allow running multiple instances of dropwatch
+
+Despite the fact that the drop_monitor module sends multicast
+notifications, it is not possible to run multiple instances of dropwatch
+since tracing of software / hardware drops can only be enabled once.
+
+Solve this by maintaining a reference count of the number of times that
+software / hardware drop tracing was enabled. Initialize the tracing
+when the reference count is incremented from 0 and tear it down when
+decremented back to 0.
+
+Note that like before, it is not possible to change configuration while
+tracing is enabled.
+
+Signed-off-by: Ido Schimmel <idosch@nvidia.com>
+---
+ net/core/drop_monitor.c | 46 ++++++++++++++++++++---------------------
+ 1 file changed, 23 insertions(+), 23 deletions(-)
+
+diff --git a/net/core/drop_monitor.c b/net/core/drop_monitor.c
+index 60d31c2feed3..29f2b8c6b8b9 100644
+--- a/net/core/drop_monitor.c
++++ b/net/core/drop_monitor.c
+@@ -47,8 +47,8 @@
+  * and the work handle that will send up
+  * netlink alerts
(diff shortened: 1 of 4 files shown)


oleksandrivantsiv (approved):

patches-sonic/0003-drop_monitor-Allow-running-multiple-instances-of-dropwatch.patch
```diff
@@ -0,0 +1,146 @@
...
++		return 0;
+ 
+ 	ops = net_dm_alert_ops_arr[net_dm_alert_mode];
+ 
+@@ -1079,7 +1077,7 @@ static int net_dm_hw_monitor_start(struct netlink_ext_ack *extack)
+ 		goto err_module_put;
+ 	}
+ 
+-	monitor_hw = true;
++	refcount_set(&refcount_hw, 1);
```
saiarcot895, line 56: This is setting the refcount to 1, but it's overwriting the increment done above?

patches-sonic/0003-drop_monitor-Allow-running-multiple-instances-of-dropwatch.patch
```diff
@@ -0,0 +1,146 @@
...
+  *
+@@ -1051,10 +1051,8 @@ static int net_dm_hw_monitor_start(struct netlink_ext_ack *extack)
+ 	const struct net_dm_alert_ops *ops;
+ 	int cpu, rc;
+ 
+-	if (monitor_hw) {
+-		NL_SET_ERR_MSG_MOD(extack, "Hardware monitoring already enabled");
+-		return -EAGAIN;
+-	}
++	if (refcount_inc_not_zero(&refcount_hw))
```
saiarcot895, line 46: Isn't this always going to return false, since the refcount starts at 0, and since the only place it's incremented is later in this function, the increment won't happen?

saiarcot895: Ignore this, somehow I thought this meant it would always return 0.

saiarcot895:
What's the reason for not upstreaming it?

Yakiv-Huryk:
> What's the reason for not upstreaming it?

This will not be accepted upstream, explnation from the patch author:

_drop monitor is both about software and hardware drops and there is no way to associate software drops with a network namespace, so these notifications continue to be sent in the initial network namespace despite netns=ok. this is fine for sonic which only cares about hardware drops._

saiarcot895 (approved):

PR #584, opened 2026-05-31
Integrate HW-MGMT Version 7.0060.1047

DannyIsa:
#### Why I did it
Integrate HW-MGMT 7.0060.1047 changes into the SONiC kernel.

#### What's in this PR

**Commit 1 — Integrate HW-MGMT 7.0060.1047 patches**
Adds the new kernel patches that come with this HW-MGMT version:
* `0060-leds-mlxreg...`: show the correct LED color in sysfs when the LED is still controlled by hardware.
* `0060-platform-mellanox-nvsw-bmc-Downstream-Add-protection...`: only let the current owner (CPU or BMC) set the I2C mux, so we don't cut off I2C traffic.
* `0061-platform-mellanox-mlxreg-io-Increase-max...`: raise the max number of mlxreg-io attributes from 128 to 250 for newer platforms.
* `0062-Add-support-for-NXP-s-PCF85053A-RTC-chip...`: add a driver for the NXP PCF85053A RTC chip.
* `0063-reset-phy-using-polling...`: reset the Realtek PHY by polling until it's done instead of just writing a register, fixing random init failures.

BMC patches (for sonic-aspeed):
* `0001-mctp-driver-net-Extend-MCTP-support...`: add new MCTP transports (SPI, NVIDIA OP-TEE VRoT) plus tracepoints, stats and test helpers.
* `0002-mctp-aspeed-IRoT-Add-initial-support...`: add NVIDIA IRoT support on Aspeed AST27xx as an MCTP transport over the HW mailbox.
* `0003-mctp-driver-net-Add-global-APIs...`: export MCTP tracepoints so transport drivers can be built as modules.
* `0004-mctp-driver-net-Fix-MCTP-over-VDM-PCIe...`: fix MCTP over VDM/PCIe so it works when built as a module.
* `0005-mctp-driver-net-Fix-MCTP-over-USB...`: fix MCTP over USB error counting and add the f_mctp USB gadget.
* `0006-jtag-Add-JTAG-core-headers...`: add JTAG core headers, GPIO mux control, single-open locking and the Aspeed internal JTAG driver.
* `0007-JTAG-Aspeed-Fix-kernel-configuration...`: wire JTAG into the kernel build so its options show up and get compiled.

Note: several patches are marked `Upstream-Status: Pending` (not upstreamed yet).

**Commit 2 — Enable BMC Kconfig for sonic-aspeed**
Turns on the kernel config options needed by the BMC patches above (MCTP transports, IRoT and JTAG) in `config.sonic-aspeed`.

**Commit 3 — Update patch commit descriptions**
Cleanup only: fill in / fix the description text of the patches. No code changes.

Signed-off-by: Daniel Isakov <disakov@nvidia.com>


abelamit (dismissed):

config.local/arm64/config.sonic-aspeed
```diff
@@ -106,5 +106,91 @@ CONFIG_CRYPTO_DEV_ASPEED=y
...
+CONFIG_I3C_HUB=m
+CONFIG_REGMAP_I3C=m
+CONFIG_SENSORS_MLXREG_FAN=m
+CONFIG_MLXREG_HOTPLUG=m
+CONFIG_MLXREG_IO=m
+CONFIG_MLX_WDT=m
+CONFIG_RTC_DRV_PCF85053A=y
+CONFIG_NVSW_BMC_HID162=m
+CONFIG_I2C_MUX_REGMAP=y
+CONFIG_PCIE_PME=n
```
paulmenzel, line 194: Please split this to a separate commit with dedicated commit message.

patches-sonic/0061-platform-mellanox-mlxreg-io-Increase-max-supported-a.patch
```diff
@@ -0,0 +1,31 @@
+Upstream-Status: Pending
```
paulmenzel, line 1: Add this in the commit message?

patches-sonic/0061-platform-mellanox-mlxreg-io-Increase-max-supported-a.patch
```diff
@@ -0,0 +1,31 @@
...
+
+From 9f11c71d770ea5a66ef3a4e6bea134aaaf62adaa Mon Sep 17 00:00:00 2001
+From: Aaron Komisar <akomisar@nvidia.com>
+Date: Thu, 28 Aug 2025 20:57:42 +0300
```
paulmenzel, line 5: Is the upstream status really pending?

paulmenzel (commented):
Thank you for the merge/pull requests. Some notes:

Bots cannot sign off anything. Please avoid this:

> Signed-off-by: sw-r2d2-bot <sw-r2d2-bot@nvidia.com>

Additionally, it’d be great if the upstream status of the commits could be more clear. I assume they have been sent for review. Could you maybe add the URLs of lore.kernel.org?

patches-sonic/0061-platform-mellanox-mlxreg-io-Increase-max-supported-a.patch
DannyIsa: Those patches are not up-streamed and are not sent for review, so marked as pending.

DannyIsa:
@paulmenzel
> Could you maybe add the URLs of lore.kernel.org?

All these patches were not sent for review yet.

patches-sonic/0061-platform-mellanox-mlxreg-io-Increase-max-supported-a.patch
paulmenzel: Strange over half a year later.

patches-sonic/0007-JTAG-Aspeed-Fix-kernel-configuration.patch
```diff
@@ -0,0 +1,37 @@
+From 7e6b94349b5fb51f0792b31cb186585f6466d8c0 Mon Sep 17 00:00:00 2001
+From: Vadim Pasternak <vadimp@nvidia.com>
+Date: Thu, 23 Apr 2026 21:40:40 +0300
+Subject: [PATCH 6.12 1/1] JTAG: Aspeed: Fix kernel configuration
+
+Change Kconfig and Makefile.
```
paulmenzel: The reasoning/motivation is missing.

config.local/arm64/config.sonic-aspeed
paulmenzel: Thank you for splitting it out. It’d be great to have a commit message, where the configuration is taken from or was generated. Several sensors are chosen for example.

patches-sonic/0004-mctp-driver-net-Fix-MCTP-over-VDM-PCIe.patch
```diff
@@ -0,0 +1,52 @@
+From ef1be57f590ab3df679fb591f22065561c796c56 Mon Sep 17 00:00:00 2001
+From: Vadim Pasternak <vadimp@nvidia.com>
+Date: Sun, 10 May 2026 14:37:56 +0300
+Subject: [PATCH 6.12 4/5] mctp: driver: net: Fix MCTP over VDM/PCIe
+
+Add fixes for MCTP over VDM/PCIe (option MCTP_TRANSPORT_PCIE_VDM).
```
paulmenzel: This is also too short and does state the problems fixed/motivation.

patches-sonic/0063-reset-phy-using-polling-function-not-register-write.patch
```diff
@@ -0,0 +1,182 @@
+From d380e09439f8576e4a89dd1f5c32850ce1c14d62 Mon Sep 17 00:00:00 2001
+From: Vadim Pasternak <vadimp@nvidia.com>
+Date: Thu, 2 Apr 2026 12:24:00 +0300
+Subject: [PATCH BMC Realtek PHY 1/1] From
```
paulmenzel: With the lines below this looks strangely formatted.

patches-sonic/0063-reset-phy-using-polling-function-not-register-write.patch
```diff
@@ -0,0 +1,182 @@
...
+ d2e5ed78b4233e73d5b392cb0ca7b61d4409e8de Mon Sep 17 00:00:00 2001 Subject:
+ [PATCH BMC Realtek PHY 1/1] reset phy using polling function not  register
+ write
+
+replace the code to reset the phy with a call to a
+function that resets the phy and waits for the reset to be done,
+preventing race conditions.
```
paulmenzel: Which exactly?

patches-sonic/0063-reset-phy-using-polling-function-not-register-write.patch
```diff
@@ -0,0 +1,182 @@
...
++	set_600mv_drive_strength(phydev);
++#endif
+ 	return 0;
+ }
+ 
+@@ -409,6 +437,49 @@ static int rtl8211f_config_init(struct phy_device *phydev)
+ 		return 0;
+ 	}
+ 
++#ifdef CONFIG_RTL_RGMII_SGMII_3809
```
paulmenzel, line 82: Use if in C and *not* preprocessor?

paulmenzel (changes requested):
Some more comments. Feel free to mark them as resolved.

patches-sonic/0063-reset-phy-using-polling-function-not-register-write.patch
DannyIsa: This is a workaround patch that we don't plan to upstream, which is exactly why it's gated behind `#ifdef CONFIG_RTL_RGMII_SGMII_3809` , the same code already ships in our OpenBMC tree.

oleksandrivantsiv (approved):

patches-sonic/0063-reset-phy-using-polling-function-not-register-write.patch
paulmenzel: It’s still in the SONiC tree and should be addressed.

yxieca:
@paulmenzel, thanks again for the detailed feedback — the points about proper sign-off and clearer upstream status are valid and important for long-term maintainability.

The author has now updated the PR description with per-patch details to improve traceability. Reworking the existing patches would require rewriting git history, which is costly at this stage, so I’m inclined to treat this as a one-time exception.
Going forward, we should hold the bar that:

- commits have proper sign-off (no bot sign-off), and
- upstream status is clearly documented (e.g., links to lore/kernel submissions where applicable).

With that context, do you feel comfortable waiving the “request changes” for this PR? If there are no other blocking concerns, I’d like to move this toward merge.

paulmenzel (commented):
I think the if in C is useful, but I won’t block anything.

The last SONiC commit should also be amended to use imperative mood: Update …. Personally, I would even squash it in the first commit. (I know SONiC squashes on merge anyway, but I do not know, what text is copied where.)

PR #589, opened 2026-06-11
Clear non-fatal errors on AER recovery failure

yurypm:
pci_aer_clear_nonfatal_status() is not called when AER recovery fails.

A PCIe switch can report an AER error with the 0000:00:00.0 AER error
source address (if the multi-error flag is not set). In this case, the
AER driver rescans the whole bus and tries to find a device reporting
the AER error.

find_source_device() is called. When the AER error source is 0,
is_error_source() returns 'true' for any device with a reported AER
error. When is_error_source() reports 'true', the AER driver checks
the e_info->multi_error_valid flag and stops iterating in
find_device_iter(). Therefore, the error is reported only for the first
device on the bus.

Then, the AER driver initiates AER recovery. If AER recovery fails,
AER error cleanup is not called for the devices. When any device on
the same bus reports a new AER error, the bus rescan process is
repeated, and the AER error is reported for the same first device with
the uncleaned AER error.

This is a side effect of an AER recovery failure. We should not be
left with stale errors. We can either perform error cleanup in the
kernel immediately after a failure, or we can leave the devices as-is
and let user-space code recover the system/PCIe tree. The current
Linux kernel leaves the devices as-is, which leads to visible side
effects in the absence of user-space monitoring.

Add a kernel boot parameter pci=aer_clear_on_recovery_failure to
enable AER error cleanup for cases where recovery fails. This prevents
stale errors from causing incorrect device identification on subsequent
AER error events.

The kernel boot parameter allows enabling this functionality on demand.
By default, it is disabled, because there could be software that
expects to see the device in an unchanged state when AER recovery
fails.

The patch was discussed with Linux maintainers here:
https://lore.kernel.org/linux-pci/CAPzpGcRCTCZtaX1EVaJNZ103THZKsoszZduY7=gwfYdcrMo-SQ@mail.gmail.com/
They prefer to fix the reason for the issue - AER recovery failure.
Proposed fix introduces new crashes in the SONiC Linux kernel during
AER error recovery and cannot be used as-is.

We are planning to enable the new functionality only for Arista
devices

### DIFF
--- a/patches-sonic/driver-arista-pci-clear-aer-error-on-recovery-failure.patch
+++ b/patches-sonic/driver-arista-pci-clear-aer-error-on-recovery-failure.patch
@@ -1,0 +1,140 @@
+From a24b3e9f6f67bed07efe06efb3cee2bf8457120a Mon Sep 17 00:00:00 2001
+From: Yury Murashka <yurypm@arista.com>
+Date: Mon, 18 May 2026 12:42:53 +0000
+Subject: [PATCH] Clear non-fatal errors on AER recovery failure
+
+pci_aer_clear_nonfatal_status() is not called when AER recovery fails.
+
+A PCIe switch can report an AER error with the 0000:00:00.0 AER error
+source address (if the multi-error flag is not set). In this case, the
+AER driver rescans the whole bus and tries to find a device reporting
+the AER error.
+
+find_source_device() is called. When the AER error source is 0,
+is_error_source() returns 'true' for any device with a reported AER
+error. When is_error_source() reports 'true', the AER driver checks
+the e_info->multi_error_valid flag and stops iterating in
+find_device_iter(). Therefore, the error is reported only for the first
+device on the bus.
+
+Then, the AER driver initiates AER recovery. If AER recovery fails,
+AER error cleanup is not called for the devices. When any device on
+the same bus reports a new AER error, the bus rescan process is
+repeated, and the AER error is reported for the same first device with
+the uncleaned AER error.
+
+This is a side effect of an AER recovery failure. We should not be
+left with stale errors. We can either perform error cleanup in the
+kernel immediately after a failure, or we can leave the devices as-is
+and let user-space code recover the system/PCIe tree. The current
+Linux kernel leaves the devices as-is, which leads to visible side
+effects in the absence of user-space monitoring.
+
+Add a kernel boot parameter pci=aer_clear_on_recovery_failure to
+enable AER error cleanup for cases where recovery fails. This prevents
+stale errors from causing incorrect device identification on subsequent
+AER error events.
+
+The kernel boot parameter allows enabling this functionality on demand.
+By default, it is disabled, because there could be software that
+expects to see the device in an unchanged state when AER recovery
+fails.
+
+The patch was discussed with Linux maintainers here:
+https://lore.kernel.org/linux-pci/CAPzpGcRCTCZtaX1EVaJNZ103THZKsoszZduY7=gwfYdcrMo-SQ@mail.gmail.com/
+They prefer to fix the reason for the issue - AER recovery failure.
+Proposed fix introduces new crashes in the SONiC Linux kernel during
+AER error recovery and cannot be used as-is.
+
+We are planning to enable the new functionality only for Arista
+devices.
+
+Signed-off-by: Yury Murashka <yurypm@arista.com>
+---
+ Documentation/admin-guide/kernel-parameters.txt |  4 ++++
+ drivers/pci/pci.c                               |  2 ++
+ drivers/pci/pci.h                               |  2 ++
+ drivers/pci/pcie/err.c                          | 13 +++++++++++++
+ 4 files changed, 21 insertions(+)
+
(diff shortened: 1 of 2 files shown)


patches-sonic/driver-arista-pci-clear-aer-error-on-recovery-failure.patch
```diff
@@ -0,0 +1,128 @@
...
+diff --git a/Documentation/admin-guide/kernel-parameters.txt b/Documentation/admin-guide/kernel-parameters.txt
+index 068891e..58ee606 100644
+--- a/Documentation/admin-guide/kernel-parameters.txt
++++ b/Documentation/admin-guide/kernel-parameters.txt
+@@ -4490,6 +4490,11 @@
+ 				enabled, this kernel boot option can be used to
+ 				disable AER error recovery when an uncorrectable
+ 				error is reported.
++		aer_clear_on_recovery_failure
++				[PCIE] If the PCIEAER kernel config parameter is
```
paulmenzel: Start on the line above?

patches-sonic/driver-arista-pci-clear-aer-error-on-recovery-failure.patch
```diff
@@ -0,0 +1,128 @@
...
+the e_info->multi_error_valid flag and stops iterating in
+find_device_iter(). Therefore, the error is reported only for the first
+device on the bus.
+
+Then, the AER driver initiates AER recovery. If AER recovery fails,
+AER error cleanup is not called for the devices. When any device on
+the same bus reports a new AER error, the bus rescan process is
+repeated, and the AER error is reported for the same first device with
+the uncleaned AER error.
+
```
paulmenzel, line 25: That sounds like the wrong behavior. Can this be fixed?

paulmenzel (commented):
I am not a PCI expert. Can you please send it upstream for review?

Reading the commit message, it sounds like a workaround, and the actual problem needs to be fixed with the stale errors? Or is it by design?

patches-sonic/driver-arista-pci-clear-aer-error-on-recovery-failure.patch
yurypm: https://lore.kernel.org/linux-pci/CAPzpGcRCTCZtaX1EVaJNZ103THZKsoszZduY7=gwfYdcrMo-SQ@mail.gmail.com/

Stale errors are a side effect of AER recovery failures and the lack of AER error cleanup following a failure. This patch aims to resolve the stale error issue.

While we could attempt to fix the AER recovery failure in one particular case, the stale error issue would persist because it is perfectly legal for drivers to reject recovery.

Linux kernel maintainers suggested fixing the AER recovery failure directly. However, the proposed fix causes kernel crashes and requires further investigation.

This patch provides an immediate fix for the stale errors observed on our devices. This allows us time to continue investigating and ultimately fix the AER recovery failure, or alternatively, disable AER recovery entirely for all Arista devices.

patches-sonic/driver-arista-pci-clear-aer-error-on-recovery-failure.patch
yurypm: This is a side effect of an AER recovery failure. We should not be left with stale errors. We can either perform error cleanup in the kernel immediately after a failure, or we can leave the devices as-is and let user-space code recover the system/PCIe tree. The current Linux kernel leaves the devices as-is, which leads to visible side effects in the absence of user-space monitoring.

patches-sonic/driver-arista-pci-clear-aer-error-on-recovery-failure.patch
yurypm: > Start on the line above?

Fixed

patches-sonic/driver-arista-pci-clear-aer-error-on-recovery-failure.patch
paulmenzel: Could you please amend the commit message to make this more clear.

patches-sonic/driver-arista-pci-clear-aer-error-on-recovery-failure.patch
paulmenzel: (Hmm, no idea how to mark threads as resolved, if they do not show up in the *Files changed* view.)

Could you please also amend the commit message with a summary of the maintainers response, and the URL to the discussion from May? (No idea why I didn’t find it, when searching.) It sounds like Lukas’ patches would solve your issue, so it would be best, if you tested these instead of having a new parameter?

yurypm: Lukas' patch introduces new crash in the kernel and doesn't work as-is.

yurypm: Added link to discussion to the patch

patches-sonic/driver-arista-pci-clear-aer-error-on-recovery-failure.patch
yurypm: Done

saiarcot895 (approved):

PR #594, opened 2026-06-23
Update AST2700 patch generation script to generate incremental patches

jack-nexthop:
Updating the AST2700 patch generation script to generate incremental patches as opposed to regenerating the entire support patch each time. Script diffs the current committed state against a fresh Aspeed upstream copy and emits only what's new upstream relative to the Aspeed patches already committed in patches-sonic/series. As before, it defaults to aspeed-master-v6.12 and can be overridden with `ASPEED_TAG`, e.g. 
```
ASPEED_TAG=v00.07.02 ./scripts/generate-aspeed-patch.sh
```
Also adds an `###-> aspeed-upstream` sub-section in the series to mark the patches the script regenerates so hand-written board patches are left untouched. Additionally adds a skip list so tool-managed blocks like `nvidia_aspeed_bmc` are skipped.

### DIFF
--- a/scripts/generate-aspeed-patch.sh
+++ b/scripts/generate-aspeed-patch.sh
@@ -1,8 +1,9 @@
 #!/bin/bash
-# Script to generate Aspeed AST2700 support patch for SONiC kernel
-# This script downloads the Aspeed kernel source and SONiC kernel source,
-# applies SONiC patches (excluding Aspeed), merges Aspeed files, and generates
-# a new patch file.
+# Generates an incremental Aspeed AST2700 patch for the SONiC kernel.
+# Diffs (kernel with all currently-committed series patches applied) against
+# (kernel with a fresh Aspeed upstream merge). The output captures whatever
+# is new in upstream relative to the existing patches-sonic/ series. See the
+# design comment further below for details.
 
 set -e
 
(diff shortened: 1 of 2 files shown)


yxieca (dismissed):

scripts/generate-aspeed-patch.sh
```diff
@@ -668,81 +877,29 @@ echo "========================================="
...
-echo "To copy to patches directory:"
-echo "  cp $OUTPUT_PATCH $KERNEL_DIR/patches-sonic/aspeed-ast2700-support-new.patch"
+echo "  # 2. Append the new patch INSIDE the existing ###-> aspeed-upstream"
+echo "  #    sub-section in patches-sonic/series (after the prior upstream"
+echo "  #    patches, before ###-> aspeed-upstream-end). It must stay ahead of"
+echo "  #    the hand-written patches, and it applies on top of the upstream"
+echo "  #    patches the diff was generated against. Example:"
+echo "  #"
+echo "  #      ###-> aspeed"
+echo "  #      ###-> aspeed-upstream"
```
saiarcot895, line 898: Does there need to be an `aspeed-upstream` line added in the `series` file?

jack-nexthop: Yes, sorry for the confusion

saiarcot895 (approved):

PR #597, opened 2026-07-09
[Action/Cherry pick][202505]rtc: armada38x: disable update interrupt …

fountzou:
Stop advertising update interrupts for the armada38x RTC driver.

SONiC switched to chrony, which enables RTC update interrupts via the RTC_UIE_ON ioctl. On the armada38x RTC driver, big skew in time can lead to pinning the CPU and triggering the RCU stall and reboot.
We clear the RTC_FEATURE_UPDATE_INTERRUPT at probe time, so RTC_UIE_ON no longer takes the native periodic-timer path and instead routes to the UIE emulation path (https://github.com/sonic-net/sonic-linux-kernel/commit/dbd69ffb55aa68b45a3db6f806ffa2a9a7d544c2).

This is a cherry-pick of https://github.com/sonic-net/sonic-linux-kernel/pull/595 for 202505.

### DIFF
--- a/patch/0018-armhf-armada38x-disable-HW-UIE.patch
+++ b/patch/0018-armhf-armada38x-disable-HW-UIE.patch
@@ -1,0 +1,81 @@
+From 73ec9b3ad05e0edc30c1b70065b92a47df3e2423 Mon Sep 17 00:00:00 2001
+From: Ioannis Fountzoulas <ioannis.fountzoulas@nokia.com>
+Date: Thu, 9 Jul 2026 04:08:20 -0400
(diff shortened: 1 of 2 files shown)


fountzou:
Hi @yejianquan,
OC test results uploaded to kusto: 5486b551-db1e-4ce9-8e59-d7cd460f3a24

Unfortunately, I don't have the permissions required to add labels. Could you please help by adding the "Tested for 202505" label? 

--Thank you

yejianquan (approved):
LGTM

PR #599, opened 2026-07-09
[Action/Cherry pick][202511]rtc: armada38x: disable update interrupt

fountzou:
Stop advertising update interrupts for the armada38x RTC driver.

SONiC switched to chrony, which enables RTC update interrupts via the RTC_UIE_ON ioctl. On the armada38x RTC driver, big skew in time can lead to pinning the CPU and triggering the RCU stall and reboot.
We clear the RTC_FEATURE_UPDATE_INTERRUPT at probe time, so RTC_UIE_ON no longer takes the native periodic-timer path and instead routes to the UIE emulation path (https://github.com/sonic-net/sonic-linux-kernel/commit/dbd69ffb55aa68b45a3db6f806ffa2a9a7d544c2).

This is a cherry-pick of https://github.com/sonic-net/sonic-linux-kernel/pull/595 for 202511.


yejianquan:
close and not cherry-pick to 202511 since 202511 is not active now

PR #603, opened 2026-07-27
Fix a bug in the Aspeed SPI driver to support full-duplex operations.

chander-nexthop:
The aspeed-txrx controller only latched MISO during its TX loop when tx_buf == rx_buf, and otherwise issued a separate post-transfer ioread8_rep burst that returns zeros. spidev always hands the driver distinct tx/rx bounce buffers, so userspace SPI_IOC_MESSAGE full-duplex reads (e.g. the FPGA register bridge) always read back zeros.

Add patches-sonic/aspeed-spi-txrx-fix-full-duplex-capture.patch to capture RX during the TX loop whenever an rx buffer is present and restrict the separate read burst to pure reads, and register it in the series after aspeed-ast2700-support.patch. Behaviour is unchanged for write-only, read-only, and shared-buffer full-duplex (e.g. the TPM) transfers.

We have given this patch to Aspeed too and they have agreed to take it in their repository in a future release. Till that happens we will need this patch.

### DIFF
--- a/patches-sonic/aspeed-spi-txrx-fix-full-duplex-capture.patch
+++ b/patches-sonic/aspeed-spi-txrx-fix-full-duplex-capture.patch
@@ -1,0 +1,47 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Chandrasekaran Swaminathan <chander@nexthop.ai>
+Date: Thu, 2 Jul 2026 00:00:00 +0000
+Subject: [PATCH] spi: aspeed-txrx: capture full-duplex RX for distinct buffers
+
+The aspeed-txrx controller implements its own transfer_one_message and
+only latches MISO during the clocked TX loop when tx_buf == rx_buf (a
+single shared buffer). For any transfer that supplies separate tx_buf
+and rx_buf pointers it skipped the in-clock capture and instead issued a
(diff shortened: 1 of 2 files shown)


yxieca (approved):
**Approve** ✅

Correct two-line fix to the vendored `spi-aspeed-txrx.c` driver. The controller only latched MISO in the TX clock loop when `tx_buf == rx_buf`; for distinct buffers it fell back to a post-transfer `ioread8_rep` burst that reads zeros (device already returned its data). Since the spidev full-duplex ABI always supplies distinct bounce buffers, userspace full-duplex reads (the FPGA register bridge) always returned zeros.

Traced all four transfer cases against the base driver source:
- write-only, pure-read, and shared-buffer full-duplex (TPM) — behaviour unchanged ✅
- distinct-buffer full-duplex (FPGA) — now captures MISO in the TX loop and no longer issues the bogus zero-read burst ✅

The two hunks are mutually consistent (no double-read for any case), and the series placement is correct — inserted after `aspeed-ast2700-support.patch` (which creates the driver) and before the b27 dts.

Non-blocking nit: the patch header uses a zeroed dummy hash / `00:00:00` date; consider regenerating with `git format-patch` for a real header. Cosmetic only.

Reviewed by Ying (with AI-assisted multi-model review).

PR #605, opened 2026-08-04
manage-config: guard secureboot config append

arista-hpandya:
Add a newline before appending the secureboot config so fragments cannot be concatenated when config.sonic lacks a trailing newline. Also factor the config paths into variables for reuse.


saiarcot895:
Does the file not currently contain a trailing newline? IMO, this is something that should be fixed in the actual sources, not worked around.

PR #606, opened 2026-08-07
Fix Y2038 netlink sequence number in libbpf

subharam-cisco:
Replace time(NULL) with a static monotonic counter in libbpf_netlink_send_recv() to avoid truncating 64-bit time_t into the 32-bit nlmsg_seq field.

### DIFF
--- a/patches-sonic/0001-libbpf-Fix-Y2038-netlink-sequence-number.patch
+++ b/patches-sonic/0001-libbpf-Fix-Y2038-netlink-sequence-number.patch
@@ -1,0 +1,48 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Subhasri-Ramasamy <subharam@cisco.com>
+Date: Thu, 6 Aug 2026 12:00:00 +0000
+Subject: [PATCH] libbpf: Fix Y2038 netlink sequence number in libbpf_netlink_send_recv
+
+Coverity CID 3439741 (Y2K38_SAFETY): libbpf_netlink_send_recv() used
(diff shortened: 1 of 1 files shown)


patches-sonic/0001-libbpf-Fix-Y2038-netlink-sequence-number.patch
```diff
@@ -0,0 +1,48 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: SONiC Build <build@sonic.net>
```
paulmenzel: Please add a person.

patches-sonic/0001-libbpf-Fix-Y2038-netlink-sequence-number.patch
```diff
@@ -0,0 +1,48 @@
...
+
+Coverity CID 3439741 (Y2K38_SAFETY): libbpf_netlink_send_recv() used
+time(NULL) to seed nlmsg_seq, implicitly truncating 64-bit time_t into
+the 32-bit nlmsg_seq field.
+
+Netlink sequence numbers only need to match responses to requests and
+should be monotonically increasing; a timestamp is unnecessary. Use a
+static counter instead and drop the unused time.h include.
+
+Signed-off-by: SONiC Build <build@sonic.net>
```
paulmenzel: Ditto.

paulmenzel (changes requested):
Please send the patch to Linux upstream for review.

saiarcot895:
Has this been submitted upstream?

PR #607, opened 2026-08-07
Cherry-pick int0002 IRQF_ONESHOT removal on top of the SCI revert

croos12:
#### Why I did it

With the ACPI SCI back on a plain `IRQF_SHARED` `request_irq()`, the INT0002 virtual GPIO
driver — which shares that IRQ line — still requests it with `IRQF_ONESHOT`. The mismatched
flags make `request_irq()` fail and `int0002_vgpio` fail to probe.

The visible symptom is a boot-time hang: after a successful ONIE install the switch freezes due to the kernel boot not progressing (~30% reproduction on SN5600). Dropping `IRQF_ONESHOT` resolves it.

#### How I did it

- Added `patches-sonic/platform-x86-int0002-remove-irqf-oneshot.patch`, backporting stable
  commit `5938fb0e3285` ("platform/x86: int0002: Remove IRQF_ONESHOT from request_irq()",
  upstream `f6bc712877f2`).
- Registered the patch in `patches-sonic/series`.

#### How to verify it

1. Build and boot a master_RC image on affected x86 hardware.
2. Confirm `int0002_vgpio` probes cleanly — `dmesg | grep -i int0002` shows no
   `IRQF_SHARED`/flags-mismatch error and no `request_irq` failure.
3. Confirm the IRQ is registered and shared: `grep int0002 /proc/interrupts`.

#### Tested branch (Please provide the tested image version)
- [x] master

#### Description for the changelog

Remove IRQF_ONESHOT from INT0002 request_irq() to fix shared-IRQ flags mismatch

PR #580, opened 2026-05-22
Revert "ACPI: OSL: Use a threaded interrupt handler for SCI"

ymd-arista:
This reverts commit [7a36b901a6eb0e9945341db71ed3c45c7721cfa9](https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/commit/?id=7a36b901a6eb).

After upgrading from Debian bookworm to trixie on modular systems, the kdump kernel started hitting a soft lockup while capturing a crash dump.  The issue is reproducible by triggering a panic in the production kernel with:

echo c | sudo tee /proc/sysrq-trigger

Once the kdump kernel boots, CPU0 gets stuck in the ACPI SCI handling path and the soft lockup watchdog eventually panics the kdump kernel, so no vmcore is produced.

The trace below was obtained by adding the following to the kdump command line: debug=1, loglevel=7, softlockup_all_cpu_backtrace=1 and softlockup_panic=1:

    watchdog: BUG: soft lockup - CPU#0 stuck for 26s! [irq/9-acpi:39]
    CPU: 0 UID: 0 PID: 39 Comm: irq/9-acpi Not tainted
      6.12.41+deb13-sonic-amd64 #1  Debian 6.12.41-1
    Hardware name: Intel Camelback Mountain CRB, BIOS
      Aboot-norcal7-7.1.6-generic-22971530 06/30/2021
    RIP: 0010:acpi_os_read_port+0x30/0xa0
    Call Trace:
     <TASK>
     acpi_hw_gpe_read+0x61/0x80
     acpi_ev_detect_gpe+0x74/0x180
     acpi_ev_gpe_detect+0xe1/0x130
     acpi_ev_sci_xrupt_handler+0x1d/0x40
     acpi_irq+0x1c/0x40
     irq_thread_fn+0x23/0x60
     irq_thread+0x1b3/0x2f0
     kthread+0xd2/0x100
     ret_from_fork+0x34/0x50
     ret_from_fork_asm+0x1a/0x30
     </TASK>
    Kernel panic - not syncing: softlockup: hung tasks

Comparing the bookworm and trixie kernels, the SCI handler was moved from a hardirq handler to a threaded handler by the commit being reverted.  Moving to a threaded IRQ regressed kdump on this hardware; reverting that commit restores the previous hardirq-based SCI handling and the kdump kernel completes the crash dump without triggering the soft lockup watchdog.

**Which release branch to backport?** 
- [ ] 202305
- [ ] 202311
- [ ] 202405
[...]

PR #618, opened 2026-08-27
Fully revert ACPI SCI Handler to Bookworm Behavior

### LINKED ISSUE #28517: Bug: ONIE install intermittently leaves DUT unbootable on master (kernel 6.12/Trixie) — boot hangs early
### Is it platform specific

mellanox

### Importance or Severity

Critical

### Description of the bug

After a successful onie-nos-install of a master (kernel 6.12.41+deb13-sonic-amd64, Debian Trixie), the DUT intermittently fails to boot into the new image. On serial console the box freezes at the GRUB hand-off and never reaches a usable kernel.

Root cause: [patches-sonic/revert-acpi-osl-use-threaded-irq-for-sci.patch](https://github.com/sonic-net/sonic-linux-kernel/commit/7113ba807ec0c01ffd86820637120cc968c46585), which reverts the upstream 6.12 change that made the ACPI SCI interrupt threaded, forcing it back to a hard IRQ and altering early-boot interrupt-servicing timing globally.  Removing the patch fixes the issue.

### Steps to Reproduce

1. Install a master/master_RC image (kernel 6.12) via ONIE (onie-nos-install /tmp/sonic-<vendor>.bin).
2. Let the DUT reboot into the installed image; watch the serial console.
3. Repeat — roughly 1 in 3 boots hangs at the GRUB kernel/initrd load.

### Actual Behavior and Expected Behavior

Actual: Boot hangs after GRUB loads the kernel/initrd

Expected: DUT boots into the installed image reliably

### Relevant log output

```shell
Booting 'SONiC-OS-master'
Loading SONiC-OS OS kernel ...
Loading SONiC-OS OS initial ramdisk ...
EFI stub: UEFI Secure Boot is enabled.
<hangs forever>
```

### Output of `show version`, `show techsupport`

```shell

```

### Attach files (if any)

_No response_

croos12:
#### Why I did it

Reverting `7a36b901a6eb` ("ACPI: OSL: Use a threaded interrupt handler for SCI") puts the ACPI SCI
back on a non-threaded handler requested with plain IRQF_SHARED. Three later upstream commits had
adjusted the surrounding code to suit the threaded, IRQF_ONESHOT SCI:

- `8e57de4307` dropped the interrupt disable from the ACPICA spin lock primitives.
- `655a6e7c0d` threaded the EC dedicated IRQ handler.
- `8f812373d1` added IRQF_ONESHOT to the INT0002 request.

Restore all three to their pre-`7a36b901a6eb` behavior, so the ACPI locking, the EC dedicated IRQ,
and the INT0002 flags match the SCI handler we restored.

##### Work item tracking
- Microsoft ADO **(number only)**: N/A

#### How I did it
- Re-land `revert-acpi-osl-use-threaded-irq-for-sci.patch` (#580) and
  `platform-x86-int0002-remove-irqf-oneshot.patch` (#607), which already carries the INT0002 side.
- Add `revert-acpi-osl-use-spin-locks-without-disabling-interrupts.patch` and
  `revert-acpi-ec-use-threaded-handler-for-dedicated-irq.patch`.
- Keep the four ordered as one block in `series`, matching upstream `master`, with a comment noting
  they are a dependent set.

#### How to verify it

1. Reboot repeatedly from ONIE Install, SONiC Reboot, and Power Cycling to confirm hang does not occur on kernel start

#### Tested branch (Please provide the tested image version)
- [x] master

#### Description for the changelog

Fully revert ACPI SCI Handler to Bookworm Behavior

### DIFF
--- a/patches-sonic/revert-acpi-ec-use-threaded-handler-for-dedicated-irq.patch
+++ b/patches-sonic/revert-acpi-ec-use-threaded-handler-for-dedicated-irq.patch
@@ -1,0 +1,45 @@
+From: Connor Roos <croos@nvidia.com>
+Date: 2026-08-05
+Subject: Revert "ACPI: EC: Use a threaded handler for dedicated IRQ"
+
+This reverts commit 655a6e7c0d83d47c36218525708c9fcfdd7f4b43.
+
+That commit moved the ACPI EC dedicated-IRQ handler to a threaded
+handler, on the grounds that commit 7a36b901a6eb ("ACPI: OSL: Use a
+threaded interrupt handler for SCI") had already made all EC code run in
+thread context on GPE-signalled systems, so the dedicated-IRQ path may as
+well match.
+
+We revert 7a36b901a6eb in
+revert-acpi-osl-use-threaded-irq-for-sci.patch, restoring hardirq
+context for the SCI.  Revert this commit too so the EC dedicated-IRQ
+path goes back to the non-threaded request_irq() handler and the EC
+interrupt handling is consistent with the pre-7a36b901a6eb behaviour.
+
+This patch and
+revert-acpi-osl-use-spin-locks-without-disabling-interrupts.patch are a
+pair: with the EC dedicated-IRQ handler back in hardirq context, ACPICA
+can once again be entered from a hardirq, so the irqsave/irqrestore
+variants that patch restores are what keep the EC path safe.  Do not
+apply one without the other, and keep both ordered after
+revert-acpi-osl-use-threaded-irq-for-sci.patch in the series.
+
+Signed-off-by: Connor Roos <croos@nvidia.com>
+---
+ drivers/acpi/ec.c | 4 ++--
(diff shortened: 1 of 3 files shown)


dgsudharsan (dismissed):

patches-sonic/revert-acpi-ec-use-threaded-handler-for-dedicated-irq.patch
```diff
@@ -0,0 +1,45 @@
+From: Connor Roos <croos@nvidia.com>
+Date: 2026-08-05
+Subject: Revert "ACPI: EC: Use a threaded handler for dedicated IRQ"
```
paulmenzel, line 3: Please use `git format-patch` formatted patches.

paulmenzel (changes requested):

saiarcot895 (approved):

dgsudharsan (approved):

PR #619, opened 2026-09-01
[patches-sonic]: Drop the bridge FDB protocol field patch

tahmed-dev:
#### Why I did it

`0002-net-bridge-vxlan-Protocol-field-in-bridge-fdb.patch` exists only to support
MAC learning through the Linux bridge FDB. With MAC synchronization moved onto the
FPM channel it is no longer needed, and carrying out-of-tree kernel patches has an
ongoing cost at every kernel rebase.

It adds an `NDA_PROTOCOL` field to the bridge/vxlan FDB and, more significantly,
makes deletes protocol-scoped so control-plane and data-plane entries can coexist.
That arbitration belongs in SONiC, not in the kernel, and nothing issues those
protocol-scoped deletes once MAC state is carried over FPM.

#### Note on NTF_EXT_VALIDATED

An earlier revision of this PR also dropped
`0003-neighbor-Add-NTF_EXT_VALIDATED-flag-for-externally-v.patch` on the grounds
that it was unused. **That was wrong and has been reverted in this PR.**

The search was for the symbol name, which has no users in FRR. These flags couple
by bit value inside a netlink policy mask, not by symbol. zebra sends
`NTF_E_MH_PEER_SYNC`, which FRR defines as `0x04`; in the kernel uapi that bit is
`NTF_EXT_EXT_VALIDATED (1 << 2)`, while `NTF_EXT_MH_PEER_SYNC` is `(1 << 3)`.
Dropping the patch narrowed `NTF_EXT_MASK` from `0x0D` to `0x09`, and
`nda_policy[NDA_FLAGS_EXT]` is `NLA_POLICY_MASK(NLA_U32, NTF_EXT_MASK)`, which
rejects out-of-mask bits with `-EINVAL`. Every `RTM_NEWNEIGH` zebra sends for a
peer-synced neighbour was refused, so the entry aged to `NUD_FAILED` and EVPN
multihoming failover broke.

Measured on a two-leaf setup, before and after the revert:

```
RTM_NEWNEIGH with NDA_FLAGS_EXT=0x04 : errno=22 (EINVAL)  ->  errno=0
peer-synced neighbour on each leaf   : FAILED             ->  REACHABLE
multihomed host to gateway during ES link failure : 20/20
```

The control message, sent without `NDA_FLAGS_EXT`, returned `errno=0` in both
cases, so only the flag bit was at issue.

#### How I did it

Remove `0002-net-bridge-vxlan-Protocol-field-in-bridge-fdb.patch` and its entry
from `patches-sonic/series`.

`0001-vxlan-bridge-Add-NDA_FLAGS_EXT-support-with-NTF_EXT_.patch` and
`0003-neighbor-Add-NTF_EXT_VALIDATED-flag-for-externally-v.patch` are both
**retained**. The series comment above `0003` now records why it must not be
dropped while on 6.12.

#### How to verify it

```
make target/debs/trixie/linux-image-6.12.41+deb13-sonic-amd64-unsigned_6.12.41-1_amd64.deb
```

The remaining series applies cleanly and the kernel builds. Verified on an Arista
7060X6-64PE-B pair, including EVPN-MH Ethernet Segment link failure.

#### Dependency / merge order

**This PR must merge LAST.** On master today, `sonic-swss` fdbsyncd still contains
the kernel FDB protocol plumbing and `sonic-buildimage` still ships the iproute2
`proto` patch that this kernel patch supports. Merging this first would remove
kernel support that shipping code still depends on.

Required to merge first:
- `sonic-swss`: drops the dead kernel FDB protocol plumbing
- `sonic-buildimage`: drops the iproute2 bridge FDB protocol field patch

#### Which release branch to backport

None requested at this time.

#### Description for the changelog

Drop the bridge FDB protocol field kernel patch.


pbrisset (commented):
Can we confirm the remaining patch stack applies cleanly and boots with only 0001-vxlan-bridge-Add-NDA_FLAGS_EXT-support-with-NTF_EXT_... left under the EVPN MH section? The important validation is that NTF_EXT_MH_PEER_SYNC still works after removing the protocol and extern_valid patches.

